PR #60, opened 2022-03-16
Update links to split GitHub repos

julien-c:
I went through the tedious task of grepping `huggingface_hub` in this repo and moving the links to their respective now-correct repos 😅

### DIFF
--- a/docs/hub/model-repos.md
+++ b/docs/hub/model-repos.md
@@ -42,7 +42,7 @@ metrics:
 ---
 ```
 
-You can find the detailed specification [here](https://github.com/huggingface/huggingface_hub/blame/main/modelcard.md).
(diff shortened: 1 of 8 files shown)


docs/hub/model-repos.md
```diff
@@ -42,15 +42,15 @@ metrics:
 ---
 ```
 
-You can find the detailed specification [here](https://github.com/huggingface/huggingface_hub/blame/main/modelcard.md).
+You can find the detailed specification [here](https://github.com/huggingface/hub-docs/blame/main/modelcard.md).
```
Pierrci, line 45: Will the modelcard be removed from `huggingface_hub`? If so, it's referenced in the course so I think it should be updated there too cc @lewtun

osanseviero (approved):
Thanks!

LysandreJik (approved):
Thanks for the effort!

adrinjalali (approved):

PR #63, opened 2022-03-17
feat: Added docTR as library filter

frgfm:
Following up on https://github.com/huggingface/huggingface_hub/pull/778, this PR adds docTR as a filter option on the hub!

Since only one task is supported for now, the snippet will be updated in the future.

Any feedback is welcome!

cc @osanseviero @Narsil


osanseviero:
This is great :rocket: :heart: 

As mentioned in original PR: we add filters when they will be useful towards users. If there is only one model, the filter does not provide much significant value for users. I think the filter will be more valuable once we have a couple of models on the Hub. Is it ok if we hold from this PR until there are a couple of models with this tag?

frgfm:
Of course @osanseviero, it makes perfect sense :+1:

PR #73, opened 2022-03-24
Add eslint

mishig25:


js/.eslintrc.cjs
```diff
@@ -0,0 +1,24 @@
+module.exports = {
```
mishig25, line 1: I've used [eslint settings from doc-builder](https://github.com/huggingface/doc-builder/blob/main/kit/.eslintrc.cjs)

please let me know if I should use [moon-landing one instead](https://github.com/huggingface/moon-landing/blob/main/server/.eslintrc.js)

js/src/lib/inferenceSnippets/serveCurl.ts
```diff
@@ -5,32 +5,39 @@ export const bodyBasic = (model: ModelData): string =>
...
+export const curlSnippetBodies: Partial<
+	Record<keyof typeof PipelineType, (model: ModelData) => string>
+> = {
 	// Same order as in js/src/lib/interfaces/Types.ts
-	"text-classification":      bodyBasic,
-	"token-classification":     bodyBasic,
+	"text-classification": bodyBasic,
+	"token-classification": bodyBasic,
 	"table-question-answering": bodyBasic,
-	"question-answering":       bodyBasic,
```
mishig25, line 17: we want these extra spaces, no?

what eslint or prettier settings would allow me to keep those spaces? I've tried using moon-landing eslint setting with no success

PR #80, opened 2022-03-25
Add image-to-image and image-generation tasks

nateraw:
Adds 2 new tasks to filter by:
- image-to-image
- image-generation

Adding these as we will encourage folks to add these tags in the [HugGAN event](https://github.com/huggingface/community-events/tree/main/huggan).

⚠️ When completing this PR, I noticed `lib/` was blocked in the .gitignore, so I removed it. Otherwise I would have to force add files. Hopefully that's ok 😅


js/src/lib/interfaces/Types.ts
```diff
@@ -36,6 +36,8 @@ export enum PipelineType {
 	"image-segmentation"                                      = "Image Segmentation",
 	"text-to-image"                                           = "Text-to-Image",
 	"image-to-text"                                           = "Image-to-Text",
+	"image-to-image"                                          = "Image-to-Image",
+	"image-generation"										  = "Image Generation",
```
osanseviero, line 40: There are some weird indentation issues here and in a couple of other sections below

osanseviero (commented):
Thanks! This is cool

* Do you have a screenshot with this built just to confirm it works and how it looks like?
* I was wondering if `image-generation`  should be called `unconditional-image-generation` to clarify that there is no input, but I was also wondering if I might be missing something

osanseviero:
cc @julien-c

js/src/lib/interfaces/Types.ts
nateraw: I blame GitHub CoPilot!! Haha I guess it inserted tabs instead of spaces, and it showed it was aligned in VSCode but looks different here. Should be taken care of now 🤗

nateraw:
As requested, here's a screenshot

Screen Shot 2022-03-29 at 12 35 47 PM

js/src/lib/interfaces/Types.ts
osanseviero: Thanks!!

osanseviero:
Thank you! Overall looks good and I think the icons are great. My only remaining question is if the task should be `image-generation` or `unconditional-image-generation` (for the cases in which the image is generated from noise)

julien-c:
not super familiar with this task, but what's the input for `image-generation`? just a seed?

osanseviero:
Usually it's a noise vector (the seed is used to have consistent noise vector for consistent output)

julien-c:
`unconditional-image-generation` sounds weird/vague to me but I don't have a better idea right now =)

julien-c:
`noise-to-image`?

osanseviero:
No strong opinion from my side, I think `noise-to-image` could work as well

julien-c:
ok let's do this unless @nateraw objects then!

julien-c:
the drawback is that for me `image-generation` is very easily confused with `text-to-image`

(Note: i don't think we want to model PWC's taxonomy too much, since ours aims to be simpler)

julien-c:
with this in mind, maybe `unconditional-image-generation` is the best one then yes =)

nateraw:
Done. Screenshot here for your viewing pleasure. 

Screen Shot 2022-03-30 at 12 44 02 PM

osanseviero (approved):
Nice,  thanks for this! :rocket:

Issue #83, opened 2022-03-30
Types restructuring


osanseviero:
I will be adding a couple of notes to track here

- [x] Change to a structure that will be easier to use/scale with granular types #84
- [x] Figure out a way to skip some types to show in datasets or models (just an extra field for this, but separate PR for cleanness) #86
- [ ] Move file to better file location
- [x] Incorporate input from datasets so we're aligned (subtasks in #86 #95) 
- [x] Automatically generate data for `datasets`

PR #86, opened 2022-03-31
Align subtasks with datasets and add field to hide a task

osanseviero:
**Second part of 4 for https://github.com/huggingface/hub-docs/issues/83**

Add subtasks based on https://github.com/huggingface/datasets/pull/4066 and add field to hide from filters at left.

Note: this does not add all tasks from https://github.com/huggingface/datasets/pull/4066, just the subtasks, since adding tasks requires some additional effort to add icons and a couple of other things. The hide will require an additional internal PR. 

cc @lhoestq


julien-c (commented):
looks good to me!

lhoestq (approved):
Great thanks :)

julien-c (approved):

julien-c:
Ok I pushed --force and rebased on top of the rebased branch =)

This should be ready to merge!

PR #89, opened 2022-04-05
add code snippet for algovera dhub

richardblythman:
You can check out the docs for our library here https://docs.algovera.ai/docs/Developer/dHub


osanseviero:
Hey there! Is my understanding correct that there are just 3 algovera models at the moment? (https://huggingface.co/AlgoveraAI)

If yes, I'm not sure adding a filter at the left is needed, since users can find all your models in https://huggingface.co/AlgoveraAI and adding a good model card with a description on how to run inference can help. We want to avoid having library tags at the left of https://huggingface.co/models that will just redirect to a couple of models.

PR #90, opened 2022-04-06
Add RL modality

mishig25:
Add `rl` to modalities list


js/src/lib/interfaces/Types.ts
```diff
@@ -75,6 +76,7 @@ export const MODALITY_LABELS: Record<Modality, string> = {
 	nlp:   "Natural Language Processing",
 	audio: "Audio",
 	cv:    "Computer Vision",
+	rl:    "Reinforcement learning",
```
osanseviero, line 79:
```suggestion
	rl:    "Reinforcement Learning",
```

osanseviero (approved):
LG!

julien-c:
maybe we should already add `tabular` too, @osanseviero ... though it's gonna conflict w/ your other PRs!

Issue #93, opened 2022-04-11
Mask token is being displayed in text classification widget


osanseviero:
See https://huggingface.co/cross-encoder/ms-marco-MiniLM-L-12-v2

maybe @mishig25 ?

PR #94, opened 2022-04-11
Fix `mask` display on Text Cls widget

mishig25:
Closes https://github.com/huggingface/hub-docs/issues/93

Before:
Screenshot 2022-04-11 at 14 46 16

After this PR:
Screenshot 2022-04-11 at 14 45 26


osanseviero (approved):
Thank you!

julien-c:
slightly off topic but text-classification is more general IMO than mask-filling so the file name should be TextClassif rather than FillMark.svelte

PR #95, opened 2022-04-12
Add all tasks from datasets

osanseviero:
Third part of 4 (or 3?) for #83

This PR adds all tasks from `tasks.json` in `datasets` with latest update from @lhoestq in https://github.com/huggingface/datasets/pull/4066 

Internal change needed: hide the new types based on `hideInModels`

Screenshot from 2022-04-12 15-41-00

Some changes along with the PR
* multimodal and time series modalities are added, although for time series we don't have yet any tasks for which we want to show models
* we moved some CV and Audio tasks to multimodal
* added a bunch of new tasks with `hideInModels` 
* changed `TASKS_DATA` and `TASKS_MODEL_LIBRARIES` to have same order as `PIPELINE_TAGS_DISPLAY_ORDER` since that really helps with diffs.

cc @lhoestq


js/src/lib/interfaces/Types.ts
```diff
@@ -310,12 +318,12 @@ export const PIPELINE_DATA = ensureRecordOfPipelines({
...
 	"text-to-speech": {
 		name:     "Text-to-Speech",
-		modality: "audio",
+		modality: "multimodal",
 		color:    "yellow",
 	},
 	"automatic-speech-recognition": {
 		name:     "Automatic Speech Recognition",
-		modality: "audio",
+		modality: "multimodal",
```
julien-c, line 326: i would leave those two in `audio` i think

js/src/lib/interfaces/Types.ts
```diff
@@ -431,16 +439,135 @@ export const PIPELINE_DATA = ensureRecordOfPipelines({
...
+		color:          "red",
 		hideInDatasets: true,
 	},
-	"structured-data-classification": {
-		name:     "Structured Data Classification",
-		modality: "other",
-		color:    "blue",
+	"tabular-classification": {
+		name:     "Tabular Classification",
+		modality: "nlp",
```
julien-c, line 449: is this supposed to be NLP? it's weird no? shouldn't it be `other` or even better a new `structured` modality?

js/src/lib/interfaces/Types.ts
```diff
@@ -1,19 +1,23 @@
...
-	audio: "Audio",
-	cv:    "Computer Vision",
-	rl:    "Reinforcement Learning",
-	other: "Other",
+	multimodal:  "Multimodal",
+	nlp:         "Natural Language Processing",
+	audio:       "Audio",
+	cv:          "Computer Vision",
+	rl:          "Reinforcement Learning",
+	time_series: "Time Series",
```
julien-c, line 19:
```suggestion
	time_series: "Time Series",
	structured:   "Structured Data",
```

julien-c: and actually I would advocate for time series to be inside structured

js/src/lib/interfaces/Types.ts
```diff
@@ -431,16 +439,135 @@ export const PIPELINE_DATA = ensureRecordOfPipelines({
...
+				name: "Fact Checking Retrieval",
+			},
+		],
+		modality:     "nlp",
+		color:        "indigo",
+		hideInModels: true,
 	},
+	"time-series-forecasting": {
+		name:     "Time Series Forecasting",
+		modality: "time_series",
```
julien-c, line 533:
```suggestion
		modality: "structured",
```

tasks/src/const.ts
```diff
@@ -6,35 +6,45 @@ import type { PipelineType } from "../../js/src/lib/interfaces/Types";
...
+	"text-retrieval":                 [],
+	/// multimodal
+	"feature-extraction":             ["sentence-transformers", "transformers"],
 	"automatic-speech-recognition":   ["espnet", "speechbrain", "transformers"],
+	"text-to-speech":                 ["espnet", "tensorflowtts"],
+	"text-to-image":                  [],
+	"image-to-text":                  [],
+	"visual-question-answering":      [],
+	"zero-shot-image-classification":  [],
+	/// audio
```
julien-c, line 34: i would order those by their order in `PIPELINE_DATA` rather than in `PIPELINE_TAGS_DISPLAY_ORDER`,  because the former is way less likely to change in the future

tasks/src/tasksData.ts
```diff
@@ -18,42 +18,51 @@ import textClassification from "./text-classification/data";
 import textGeneration from "./text-generation/data";
 
 
-// To make comparisons easier, task order is the same as in /lib/interfaces/Types.ts
+// To make comparisons easier, task order is the same as in PIPELINE_TAGS_DISPLAY_ORDER in /lib/interfaces/Types.ts
```
julien-c, line 21:
```suggestion
// To make comparisons easier, task order is the same as in PIPELINE_DATA in /lib/interfaces/Types.ts
```
same suggestion here

tasks/src/const.ts
julien-c: (or just alphabetically if it's simpler)

js/src/lib/interfaces/Types.ts
lhoestq: Having a `structured` modality sounds good to me

js/src/lib/interfaces/Types.ts
lhoestq: Maybe the number of Time Series datasets/models is so small that it's maybe overkill to have a dedicated modality yet.

However in the long run I expect it to become separated to the other structured datasets we'll have. We may want to have a separate modality just for time series at one point anyway for classification, forecasting, anomaly detection, etc. As vision, audio and text modalities, time series require very specific preprocessing and model architectures.

Though I agree such datasets often come with structured metadata to help predictions

js/src/lib/interfaces/Types.ts
osanseviero: Yes, I agree and actually had a comment for exactly this in the `datasets` PR. Will update accordingly

osanseviero: Although should modality be `tabular` for alignment with AutoTrain I think

tasks/src/const.ts
osanseviero: Let's go with alphabetically so it keeps us sane :)

js/src/lib/interfaces/Types.ts
julien-c: `tabular` if we don't include time-series inside of them, `structured` if we do?

js/src/lib/interfaces/Types.ts
julien-c: individual tasks will still time-series related so I like using `structured` as an umbrella for a few different things, personally. Feels like a better level of generality

osanseviero: What about table tasks? Technically it's also structured no? But we've kept them under NLP

osanseviero: Since modality is not heavily used anywhere, I feel it's safe to change afterward if needed once we see more adoption. Let's go with `structured` and have everything together, so we don't have sections just pointing to a couple of models/datasets. Later on we can always split pragmatically.

PR #97, opened 2022-04-12
Add support for NeMo toolkit

titu1994:
# Changelog
- Add support for NVIDIA NeMo toolkit (ASR domain only for now).

### DIFF
--- a/js/src/lib/interfaces/Libraries.ts
+++ b/js/src/lib/interfaces/Libraries.ts
@@ -23,6 +23,7 @@ export enum ModelLibrary {
 	"stanza"                 = "Stanza",
 	"fasttext"               = "fastText",
 	"stable-baselines3"      = "Stable-Baselines3",
+	"nemo"					 = "NeMo",
 }
 
 export const ALL_MODEL_LIBRARY_KEYS = Object.keys(ModelLibrary) as (keyof typeof ModelLibrary)[];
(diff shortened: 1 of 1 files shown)


js/src/lib/interfaces/Libraries.ts
```diff
@@ -314,6 +315,38 @@ checkpoint = load_from_hub(
...
+const nemoDomainResolver = (domain: string, model: ModelData) => {
+	var model_name = `${(model.id).split("/").pop()}.nemo`;
+
+	switch (domain) {
+		case "ASR":
+		    return `import nemo.collections.asr as nemo_asr
+		    from huggingface_hub import hf_hub_download
+		    
+		    path = hf_hub_download(repo_id="${model.id}",filename="${model_name}")
+		    asr_model = nemo_asr.models.ASRModel.restore_from(path)
```
julien-c, line 327: would you be willing to add `huggingface_hub` as a dependency to `nemo` going forward so that the syntax is as simple as possible, e.g.

```python
import nemo.collections.asr as nemo_asr

asr_model = nemo_asr.models.ASRModel.from_pretrained("${model.id}")
```
?

js/src/lib/interfaces/Libraries.ts
```diff
@@ -314,6 +315,38 @@ checkpoint = load_from_hub(
...
+
+			transcriptions = asr_model.transcribe(["file.wav"])
+		    `;
+		default:
+			return undefined;
+	}
+};
+
+const nemo = (model: ModelData) => {
+	var command = undefined;
```
julien-c, line 337:
```suggestion
	let command = undefined;
```

julien-c (commented):
left a few comments, other than that, looks good to me

js/src/lib/interfaces/Libraries.ts
titu1994: Yes that's the plan. A later Nemo point release will have internal support for from_pretrained with huggingface_hub so that users can seamlessly use both Nemo and their own / other public uploaded Nemo models using just one line of from_pretrained

titu1994: Right now we are prototyping this implementation since we are already at code freeze for the current branch, but we expect the point release to be trivial to implement (so soon after the prototype is ready and working, we can finalize the internal operations in Nemo)

julien-c (approved):

PR #98, opened 2022-04-13
Shuffle tasks display a little bit

julien-c:
As a quick fix to become "less-NLP-centric" before we revamp the /models (and /datasets) pages to show a modality selector front and center (cc @gary149)

### Screenshots

before click on view more:
Screenshot 2022-04-13 at 15 46 45

full view:
image


osanseviero:
I would prioritize tasks that will lead to many models. E.g. image classification > object detection, ASR > audio classification, etc.

julien-c:
that shuffle was really a random one.

my point is that we can maybe consider _not_ ordering by popularity but showcasing a more diverse set of tasks and not be too NLP-heavy

but not strong opinion on the actual ordering!

osanseviero:
Yes, fully agree with the intention of not ordering by popularity. I think it's important though that the filters we show do lead to a sensible number of models. If we give a filter for only <10 models, it seems meh as a user.

gary149:
Maybe you can edit and try a "better" ordering @osanseviero ?

osanseviero:
I just tried a change...maybe we should change some colors, but I think this set of tasks leads to a good number of models and shows some diversity in tasks we support

tasks_test

julien-c:
that works for me!

gary149 (approved):

PR #115, opened 2022-04-20
Remove token from inference widget

allendorf:


coyotte508:
All the calls like:

```ts
<InferenceWidget
							hydrate-props={{
								apiUrl: apiInferenceUrl,
								apiToken: auth?.u.token ?? "",
								callApiOnMount: true,
								model,
								noTitle: true,
							}}
						/>
```

need to have the `apiToken` prop removed. So that `auth.u.token` is not passed to client-side frontend code.

coyotte508 (approved):
Disregard my previous comment! LGTM

Pierrci (commented):
Actually, I think we can leave this code untouched - `apiToken` is `undefined` by default anyway so it shouldn't affect our switch to cookies on the website (and you rightly removed everywhere we were passing the session bearer token in https://github.com/huggingface/moon-landing/pull/2712).

But this code and the widgets could also be used outside of the website with a regular user access token - in which case being able to pass it to `apiToken` still makes sense.

allendorf:
works for me!

PR #116, opened 2022-04-21
Adjust link to TF notebook on Question Answering

NimaBoscarino:
A link in the Question Answering task page points to an incorrect notebook. This PR fixes that link.

Note: All links that point to `huggingface/notebooks/blob/master/...` already redirect to `huggingface/notebooks/blob/main/...`, but I can change all `master` to `main` while I'm working through the revamp.


stevhliu (approved):
Thanks for catching this!

PR #119, opened 2022-04-26
Set `credentials: include` in InferenceWidget

allendorf:
- Part of https://github.com/huggingface/moon-landing/issues/2758 
- Tied to https://github.com/huggingface/moon-landing/pull/2772

### DIFF
--- a/js/src/lib/components/InferenceWidget/shared/helpers.ts
+++ b/js/src/lib/components/InferenceWidget/shared/helpers.ts
@@ -90,7 +90,7 @@ async function callApi(
 			method: "POST",
 			body,
 			headers,
-			credentials: includeCredentials,
+			credentials: includeCredentials ? "include" : null,
 		}
 	);
 }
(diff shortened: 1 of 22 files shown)


julien-c:
are there really cases where we don't want to `include credentials`?

@Pierrci could be true by default no?

Pierrci:
@julien-c having it `false` by default is equivalent in moon-landing to simply replacing the token that was passed explicitly by this new param. Given that one of the initial goals for open-sourcing the widgets was to have an npm package at some point I would prefer to keep it this way? (since this param will probably only ever be useful to hf.co)

js/src/lib/components/InferenceWidget/shared/helpers.ts
```diff
@@ -89,6 +90,7 @@ async function callApi(
 			method: "POST",
 			body,
 			headers,
+			credentials: includeCredentials ? "include" : null,
```
Pierrci, line 93:
```suggestion
			credentials: includeCredentials ? "include" : "same-origin",
```
^^ let's use the default value rather

PR #122, opened 2022-04-27
Added type for multi-input text classification

nrajani:
Based on discussion with @lhoestq to add support for text classification that take in multiple inputs.


lhoestq:
Let me know if it sounds good to you @osanseviero, this is a subtask of text-classification when it has several texts as input (for similarity or NLI for example). This will be useful for autoeval and autotrain

The associated PR in `datasets` is here: https://github.com/huggingface/datasets/pull/4244

osanseviero (approved):
This makes sense, thanks for the PR!

PR #64, opened 2022-03-17
Add the SegFormer blog post to Image Segmentation Task

segments-tobias:
Today, a blog post on how to "[Fine-Tune a Semantic Segmentation Model with a Custom Dataset](https://huggingface.co/blog/fine-tune-segformer)" was released on the HF blog (see [PR #234](https://github.com/huggingface/blog/pull/234)).

This PR links the blog post in the "Useful Resources" section of the Image Segmentation task.
It also adds the `segments/sidewalk-semantic` dataset and `segments-tobias/segformer-b3-finetuned-segments-sidewalk` model as an example.

PR #124, opened 2022-04-28
Fix segmentation task resources text

segments-tobias:
In my [previous PR](https://github.com/huggingface/hub-docs/pull/64), I added a Resources section to the image segmentation task. However, I just noticed I copied the text from token classification without changing it. This PR fixes that issue.


osanseviero (approved):
LG, thanks!

PR #125, opened 2022-04-28
Update tags order

NielsRogge:
This PR updates the order in which "unconditional image generation" and "image segmentation" are displayed.

As the latter is more mature, has more scripts and models, it makes sense to show this one rather than unconditional image generation.


osanseviero (approved):
LGTM cc @julien-c

PR #126, opened 2022-04-29
Update NeMo ASR resolver

titu1994:
# Changelog
- Updates the ASR from_pretrained() for NeMo 1.8.2+ to directly use the provided model.id


osanseviero (approved):
Super nice! 🎉Thank you!

osanseviero (approved):
I think you can remove the modelName variable now

titu1994:
I'd like to keep the model_name variable - in the future if we have speaker diarization, its a bit more involved than just calling from_pretrained() with the model.id

julien-c (approved):

PR #135, opened 2022-05-08
Add NPC Engine library tag to Hub

eublefar:
NPC Engine is local ONNX server for serving models for game engines (it uses DirectML ONNX provider, can be built/shipped with a game and provide inference with GPU acceleration with no CUDA dependency on windows).


osanseviero:
Hey there! This tag will be automatically detected to all models that have the `npc-engine` tag as a library tag, but I don't see any models with that tag https://huggingface.co/models?other=npc-engine. I think it might make more sense to add some repos with that tag before adding a filter button that will map to 0 models.

eublefar:
Oh, sorry, my bad. For some reason I thought I couldn't just put anything I want into the library tags field, but I see now that in fact I should've. Give me a few days and I'll tag all the related models. 

I'll close the commit and reopen when ready.

PR #136, opened 2022-05-10
Add metadata fields to model-index for reproducibility

lewtun:
This PR adds new fields to the `model-index` object of model cards to enable fine-grained reproducibility / comparison with evaluation results. This will also allow us to provide fine-grained filters on leaderboards so that people can distinguish e.g. evaluations on the validation vs test sets, different configs etc.

Note that I've added a `config` field as my understanding is that people currently use the `name` field as a free text entry that is displayed on the frontend.

As discussed with the Papers with Code team, adding new fields shouldn't impact the integration.

Internal Slack threads: 

* https://huggingface.slack.com/archives/C035S5G2J3D/p1649837714080049
* https://huggingface.slack.com/archives/C01PNHFP214/p1651585598704829
* https://huggingface.slack.com/archives/C01PNHFP214/p1651677740092189

This PR is a copy of https://github.com/huggingface/huggingface_hub/pull/838 to account for the move from `huggingface_hub` to `hub-docs`


modelcard.md
```diff
@@ -2,7 +2,8 @@
 language:
 - {lang_0}  # Example: fr
 - {lang_1}  # Example: en
-license: {license}  # Example: apache-2.0 or any license from https://huggingface.co/docs/hub/model-repos#list-of-license-identifiers
+license: {license}  # Example: apache-2.0 or any license from https://hf.co/docs/hub/model-repos#list-of-license-identifiers
+library_name: {library_name}  # Example: keras or any library from https://github.com/huggingface/huggingface_hub/blob/main/js/src/lib/interfaces/Libraries.ts
```
julien-c, line 6:
```suggestion
library_name: {library_name}  # Optional. Example: keras or any library from https://github.com/huggingface/huggingface_hub/blob/main/js/src/lib/interfaces/Libraries.ts
```

julien-c (approved):
thanks! looks good

modelcard.md
```diff
@@ -13,23 +14,29 @@ datasets:
...
+  - task:
+      type: {task_type}             # Required. Example: automatic-speech-recognition
+      name: {task_name}             # Optional. Example: Speech Recognition
     dataset:
-      type: {dataset_type}  # Required. Example: common_voice. Use dataset id from https://hf.co/datasets
-      name: {dataset_name}  # Optional. Example: Common Voice zh-CN
-      args: {arg_0}         # Optional. Example: zh-CN
+      type: {dataset_type}          # Required. Example: common_voice. Use dataset id from https://hf.co/datasets
+      name: {dataset_name}          # Required. Example: Common Voice (French)
+      config: {dataset_config}      # Optional. Example: fr
```
osanseviero, line 27: Do we have any doc that documents what config is? I think that would be useful

modelcard.md
```diff
@@ -13,23 +14,29 @@ datasets:
...
+      split: {dataset_split}        # Optional. Example: test
+      revision: {dataset_revision}  # Optional. Example: 5503434ddd753f426f4b38109466949a1217c2bb
+      args:
+        {arg_0}: {value_0}          # Optional. Additional arguments to `load_dataset()`. Example for wikipedia: language: en
+        {arg_1}: {value_1}          # Optional. Example for wikipedia: date: 20220301
```
osanseviero, line 32: Do we assume all arguments are either string or number, or can it also be that an argument is a nested dict or list, etc?

modelcard.md
```diff
@@ -13,23 +14,29 @@ datasets:
...
     metrics:
-      - type: {metric_type}    # Required. Example: wer
-        value: {metric_value}  # Required. Example: 20.90
-        name: {metric_name}    # Optional. Example: Test WER
-        args: {arg_0}          # Optional. Example for BLEU: max_order
+      - type: {metric_type}         # Required. Example: wer. Use metric id from https://hf.co/metrics
+        value: {metric_value}       # Required. Example: 20.90
+        name: {metric_name}         # Optional. Example: Test WER
+        args: {arg_0}               # Optional. Example for BLEU: max_order
```
osanseviero, line 37: Should this also be `{arg_0}: {value_0}`

osanseviero (approved):
Overall looks good since it was already approved in the original PR, I left a couple of minor questions

modelcard.md
julien-c: (i think?)

osanseviero: Second param is `data_dir` I think https://huggingface.co/docs/datasets/master/en/package_reference/loading_methods#datasets.load_dataset

julien-c: hmm no `name` no?

> name (str, optional) — Defining the name of the dataset configuration

pinging @lhoestq for guidance

modelcard.md
lewtun: Good catch!

modelcard.md
```diff
@@ -13,23 +14,31 @@ datasets:
...
+  - task:
+      type: {task_type}             # Required. Example: automatic-speech-recognition
+      name: {task_name}             # Optional. Example: Speech Recognition
     dataset:
-      type: {dataset_type}  # Required. Example: common_voice. Use dataset id from https://hf.co/datasets
-      name: {dataset_name}  # Optional. Example: Common Voice zh-CN
-      args: {arg_0}         # Optional. Example: zh-CN
+      type: {dataset_type}          # Required. Example: common_voice. Use dataset id from https://hf.co/datasets
+      name: {dataset_name}          # Required. A pretty name for the dataset. Example: Common Voice (French)
+      config: {dataset_config}      # Optional. The name of the dataset configuration used in `load_dataset()`. Example: fr in `load_dataset("common_voice", "fr")`. See the `datasets` docs for more info: https://huggingface.co/docs/datasets/package_reference/loading_methods#datasets.load_dataset.name
```
lewtun, line 27: Added an example and link to the `datasets` docs here

modelcard.md
```diff
@@ -13,23 +14,31 @@ datasets:
...
+        {arg_1}: {value_1}          # Optional. Example for wikipedia: date: 20220301
     metrics:
-      - type: {metric_type}    # Required. Example: wer
-        value: {metric_value}  # Required. Example: 20.90
-        name: {metric_name}    # Optional. Example: Test WER
-        args: {arg_0}          # Optional. Example for BLEU: max_order
+      - type: {metric_type}         # Required. Example: wer. Use metric id from https://hf.co/metrics
+        value: {metric_value}       # Required. Example: 20.90
+        name: {metric_name}         # Optional. Example: Test WER
+        config: {metric_config}     # Optional. The name of the metric configuration used in `load_metric()`. Example: bleurt-large-512 in `load_metric("bleurt", "bleurt-large-512")`. See the `datasets` docs for more info: https://huggingface.co/docs/datasets/v2.1.0/en/loading#load-configurations
```
lewtun, line 37: @lvwerra pointed out to me that metrics also have configs like datasets, so I've also added this field here

modelcard.md
julien-c: pretty sure that we also use (or used to use) the `config` name interchangeably on datasets side though

no?

lewtun: Yeah I think in `datasets` we use `name` in the `load_dataset()` and `load_metric()` functions, but `config_name` in `DatasetInfo` and `MetricInfo`. The former is possibly legacy from TFDS: https://www.tensorflow.org/datasets/api_docs/python/tfds/load

lhoestq: `name` corresponds to the config name in `load_dataset` indeed.

And "pretty name" in the YAML tags is the name of the full dataset. Here in `name`you may specify the pretty name of the subset (per language for example)

PR #160, opened 2022-05-30
Fix widgettextarea

mishig25:
Copying changes from [huggingface_hub#625](https://github.com/huggingface/huggingface_hub/pull/625)

widget files were migrated to `hub-docs` repository on March 16, as can be seen [here](https://github.com/huggingface/hub-docs/commit/33b0b76b654c02a7755feeea837a9c2f05538615), which is after [huggingface_hub#625](https://github.com/huggingface/huggingface_hub/pull/625) was merged. Why [huggingface_hub#625](https://github.com/huggingface/huggingface_hub/pull/625) would not be present in `hub-docs` from the beginning, @LysandreJik @julien-c ?

### DIFF
--- a/js/src/lib/components/InferenceWidget/shared/WidgetTextarea/WidgetTextarea.svelte
+++ b/js/src/lib/components/InferenceWidget/shared/WidgetTextarea/WidgetTextarea.svelte
@@ -3,8 +3,8 @@
 
 	import WidgetLabel from "../WidgetLabel/WidgetLabel.svelte";
 
-	export let label: string = "";
-	export let placeholder: string = "Your sentence here...";
+	export let label = "";
+	export let placeholder = "Your sentence here...";
 	export let value: string;
 
 	let textAreaEl: HTMLTextAreaElement;
--- a/js/src/lib/components/InferenceWidget/shared/WidgetTextarea/WidgetTextarea.svelte
+++ b/js/src/lib/components/InferenceWidget/shared/WidgetTextarea/WidgetTextarea.svelte
@@ -1,43 +1,40 @@
-<script lang="ts">
+<script>
+	import { tick } from "svelte";
+
 	import WidgetLabel from "../WidgetLabel/WidgetLabel.svelte";
 
 	export let label: string = "";
 	export let placeholder: string = "Your sentence here...";
 	export let value: string;
 
-	// hack to handle FireFox contenteditable bug
-	let innterHTML: string;
-	let spanEl: HTMLSpanElement;
-	const REGEX_SPAN = /<span .+>(.*)<\/span>/gms;
-	$: {
-		if (spanEl && innterHTML && REGEX_SPAN.test(innterHTML)) {
-			innterHTML = innterHTML.replace(REGEX_SPAN, (_, txt) => {
-				return txt;
-			});
-			spanEl.blur();
+	let textAreaEl: HTMLTextAreaElement;
+
+	const HEIGHT_LIMIT = 500 as const;
+
+	async function resize() {
+		if (!!textAreaEl) {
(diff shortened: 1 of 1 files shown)


js/src/lib/components/InferenceWidget/shared/WidgetTextarea/WidgetTextarea.svelte
```diff
@@ -1,43 +1,40 @@
...
+	async function resize() {
+		if (!!textAreaEl) {
+			textAreaEl.style.height = "0px";
+			await tick();
+			textAreaEl.style.height =
+				Math.min(textAreaEl.scrollHeight, HEIGHT_LIMIT) + "px";
 		}
 	}
+
+	$: {
```
beurkinger, line 23: Is value only there to trigger the re-compute ? Maybe `beforeUpdate` or `afterUpdate` could be used in this case ?

beurkinger: Interesting ! I don't know if Svelte provides a way to compare old and new props on refresh, like React does ? I don't know what is the best way of handling this kind of behavior. Did you find this trick in the doc or did you come up with it ?

mishig25: I guess more or less inspired from https://svelte.dev/tutorial/reactive-statements

pngwn: `beforeUpdate` and `afterUpdate` will run when _any_ prop changes, so you'll probably end up with unnecessary calls in that case. Reactive declarations are preferred for granular updates/ calls. 

The general rule is you need to list the dependency for the reaction declaration somewhere inside the reactive declaration. The standard recommendation is to make the dep an argument to the function where possible `$: resize(value)` but in this case this function is a side-effect (you could pass the arg in and not use it but you may incur the wrath of the linting or typescript gods).

Simplest version of the current code is just `$: value, resize();`

However, I'd argue that a reactive declaration isn't the best way to achieve this. Ideally you'd bind this to a more meaningful trigger rather than the updating value. [An action](https://svelte.dev/tutorial/actions) is ideal for this because it allows you to respond to the event easily, cleanup afterwards and doesn't require manually listing dependencies (which is always confusing without context).

We have similar behaviour in a Gradio component, we need to handle some other options (and the code really needs cleaning up) but you can see how you might use an action to achieve this here: https://github.com/gradio-app/gradio/blob/main/ui/packages/form/src/Textbox.svelte#L38-L78. We are simply attaching some listeners, ensuring the DOM has updated, and limiting the size. We then remove the listeners when the component is destroyed. Events are the most appropriate trigger here imo.

The reason this makes more sense as an action is because the behaviour and effects are entirely limited to a single element. Listeners and style updates on the element itself are all that are required. This could be externalised as a utility if needed because actions are self contained, even though they interact directly with elements.

mishig25: @pngwn thanks so much for an extensive answer!

beurkinger: thanks for sharing your knowledge @pngwn, I already feel much wiser 🧙‍♂️

LysandreJik:
> Why https://github.com/huggingface/huggingface_hub/pull/625 would not be present in hub-docs from the beginning, 

Very likely an error! Thanks for fixing that, @mishig25!

mishig25:
Superceded by https://github.com/huggingface/hub-docs/pull/214

PR #163, opened 2022-05-31
High level pass for new docs

osanseviero:


osanseviero:
@NimaBoscarino one section missing in repositories-getting-started is "Adding files to a repository (Web UI)". Would you be able to give a hand with that one?

PR #169, opened 2022-06-02
git co -> git checkout

coyotte508:
`git co` is a shortcut not supported by everyone (eg default git on linux doesn't have it)


julien-c:
yep good catch

PR #177, opened 2022-06-03
Added Stanza snippet

merveenoyan:


merveenoyan:
Please do not review now :)

js/src/lib/interfaces/Libraries.ts
```diff
@@ -239,6 +239,14 @@ nlp = spacy.load("${nameWithoutNamespace(model.id)}")
 import ${nameWithoutNamespace(model.id)}
 nlp = ${nameWithoutNamespace(model.id)}.load()`;
 
+const stanza = (model: ModelData) =>
+	`from stanza import Pipeline
+
+stanza.download("${nameWithoutNamespace(model.id)}")
+nlp = stanza.Pipeline("${nameWithoutNamespace(model.id)}")
```
osanseviero, line 246: In stanza you just need the language, so you can also remove `stanza-` from the id. 

```
import stanza
stanza.download('en')
nlp = stanza.Pipeline('en')

File: js/src/lib/interfaces/Libraries.ts [hunk 2]
```diff
@@ -239,6 +239,14 @@ nlp = spacy.load("${nameWithoutNamespace(model.id)}")
 import ${nameWithoutNamespace(model.id)}
 nlp = ${nameWithoutNamespace(model.id)}.load()`;
 
+const stanza = (model: ModelData) =>
+	`from stanza import Pipeline
+
+stanza.download("${nameWithoutNamespace(model.id)}")
+nlp = stanza.Pipeline("${nameWithoutNamespace(model.id)}") 
+doc = nlp("My name is Omar and I live in Zürich.")`;
```
osanseviero, line 247: I think we should remove this as the input is dependant on the language

js/src/lib/interfaces/Libraries.ts
```diff
@@ -426,6 +434,12 @@ export const MODEL_LIBRARIES_UI_ELEMENTS: { [key in keyof typeof ModelLibrary]?:
 		repoUrl:  "https://github.com/speechbrain/speechbrain",
 		snippet:  speechbrain,
 	},
+	"stanza": {
+		btnLabel: "stanza",
```
osanseviero, line 438:
```suggestion
		btnLabel: "Stanza",
```

osanseviero (changes requested):
Thanks for the PR! :fire: :fire: :fire: 

There are a couple of things we need to change for this to work :smile:

merveenoyan:
latest-est version:
Ekran Resmi 2022-06-03 15 36 23

js/src/lib/interfaces/Libraries.ts
merveenoyan: see below :)

js/src/lib/interfaces/Libraries.ts
```diff
@@ -239,6 +239,13 @@ nlp = spacy.load("${nameWithoutNamespace(model.id)}")
 import ${nameWithoutNamespace(model.id)}
 nlp = ${nameWithoutNamespace(model.id)}.load()`;
 
+const stanza = (model: ModelData) =>
+	`from stanza import Pipeline
```
osanseviero, line 243: The current import will not work for lines 245 and 256 which use `stanza`
```suggestion
	`import stanza
```

osanseviero (approved):
Left a final comment but otherwise looks good! :fire:

julien-c:
very cool!

PR #180, opened 2022-06-04
Adjust paths, add widget examples page

NimaBoscarino:
Some paths needed to be updated, due to the revamp. Also, the Widget Examples page had been omitted from the revamp by mistake.

### DIFF
--- a/docs/hub/models-widgets-examples.md
+++ b/docs/hub/models-widgets-examples.md
@@ -1,0 +1,303 @@
+# Widget Examples
+
+## Natural Language Processing
+
+### Fill-Mask
+
+```yaml
+widget:
+- text: "Paris is the <mask> of France."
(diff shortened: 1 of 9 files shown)


mishig25 (approved):
thanks a lot for this, lgtm!

docs/hub/models-widgets-examples.md
```diff
@@ -0,0 +1,303 @@
+# Widget Examples
```
osanseviero, line 1: Could we maybe do separate page for each widget example, and in the main page have an index similar to the one in http://hf.co/docs/hub, one per modality? I think that would be cooler and goes along our idea of split pages

osanseviero (commented):
Looks very nice, although I think having separate pages on this would be nice as well!

PR #183, opened 2022-06-05
[Spaces] Configuration reference newlines

cbensimon:
Actually render new lines in the hub documentation

Before :
Screenshot 2022-06-05 at 18 53 09

After :
Screenshot 2022-06-05 at 18 52 37

### DIFF
--- a/docs/hub/spaces-config-reference.md
+++ b/docs/hub/spaces-config-reference.md
@@ -2,41 +2,41 @@
 
 Spaces are configured through the `YAML` block at the top of the **README.md** file at the root of the repository. All the accepted parameters are listed below.
 
-**`title`** : _string_
-Display title for the Space.
+**`title`** : _string_  
+Display title for the Space.  
 
-**`emoji`** : _string_
-Space emoji (emoji-only character allowed).
+**`emoji`** : _string_  
+Space emoji (emoji-only character allowed).  
 
-**`colorFrom`** : _string_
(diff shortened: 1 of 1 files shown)


docs/hub/spaces-config-reference.md
```diff
@@ -2,41 +2,41 @@
 
 Spaces are configured through the `YAML` block at the top of the **README.md** file at the root of the repository. All the accepted parameters are listed below.
 
-**`title`** : _string_
-Display title for the Space.
+**`title`** : _string_  
+Display title for the Space.
```
julien-c, line 6: Not sure what happens here.
when adding spaces at the end of lines, they'll be rendered as a newline? Maybe it would be more explicit to add a `<br>` or just to add an extra newline between the two lines no?

cbensimon: See https://daringfireball.net/projects/markdown/syntax#p

As you like but I think that a `<br>` tag will be a bit noisy and make the Markdown / source code version of the doc less readable

Adding an extra newline would create a full paragraph and it's not the intent here

Additional spaces at the end of the blocks have been added so future diffs are cleaner (just like trailing commas)

julien-c: Your solution works for me, but you should add a `<!-- comment -->` for other team members not to delete the trailing whitespace =)

osanseviero (approved):
Looks good, thanks!

PR #185, opened 2022-06-06
[Spaces Changelog] Fix formatting so lists render properly

NimaBoscarino:
Closes https://github.com/huggingface/moon-landing/issues/3150

If there's no blank line between a heading (e.g. `##`) and a paragraph, the formatted content in the paragraph won't render right. That can mean reformatted text with backticks doesn't render, or in this case the \<li\> elements don't get generated.


mishig25 (approved):
lgtm!

osanseviero (approved):
Thanks! LG!

PR #187, opened 2022-06-07
📝 Add documentation about gated models

SBrandeis:


julien-c (commented):
looks good to me, thanks for writing this

NimaBoscarino (approved):
Looks great! Before merging, would it be possible to change the screenshots? Right now they say "dataset" in them. (I'm setting up Moon Landing, so if you want I can do this too)

PR #188, opened 2022-06-07
Bring back PR advanced guide

mishig25:
https://huggingface.co/docs/hub/repositories-pull-requests-advanced was removed accidentally?

Bringing it back at https://moon-ci-docs.huggingface.co/docs/hub/pr_188/en/repositories-pull-requests-advanced

### DIFF
--- a/docs/hub/_toctree.yml
+++ b/docs/hub/_toctree.yml
@@ -10,6 +10,9 @@
     title: Repository Settings
   - local: repositories-pull-requests-discussions
     title: Pull Requests & Discussions
+    sections:
+    - local: repositories-pull-requests-advanced
+      title: Advanced Usage
   - local: repositories-next-steps
     title: Next Steps    
   - local: repositories-licenses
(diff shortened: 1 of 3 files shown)


osanseviero (commented):
In one of the previous reviews I think someone suggested to not have an advanced part. The content is in https://huggingface.co/docs/hub/repositories-pull-requests-discussions already.

If we do split up, I would keep the "How do I manage Pull requests locally?" in the not advanced. I'm :+1: for moving the advanced usage to a separate page

julien-c:
i have no strong opinion on the question

mishig25:
> In one of the previous reviews I think someone suggested to not have an advanced part. The content is in https://huggingface.co/docs/hub/repositories-pull-requests-discussions already

oh I see, that is why the page was removed

I'm happy to keep https://huggingface.co/docs/hub/repositories-pull-requests-discussions  as it is right now without splitting

PR #192, opened 2022-06-09
Tweak dropzone JS syntax

julien-c:
i don't remember if this works in all browsers but it should no?


gary149 (approved):
Yes, it does work on all browsers.

PR #200, opened 2022-06-16
update links

kamalkraj:
Update links in sagemaker inference docs


julien-c:
hmm yeah I don't think there's a standard spec, we're using out of the box `marked` for this if I'm not mistaken

Note that the first link at least was probably also broken on GitHub

julien-c:
(stale)

PR #203, opened 2022-06-21
Fix `lfs-enable-largefiles` command in `Getting Started with Repositories` docs

mariosasko:
Pass the current directory as a path in the `huggingface-cli lfs-enable-largefiles` command to avoid an error due to missing command args.


SBrandeis (approved):
Good catch! Thanks!

PR #217, opened 2022-06-28
Add binary classification to Types.ts

lewtun:
This PR is a precursor to https://github.com/huggingface/datasets/pull/4584#pullrequestreview-1021298989 and includes a `binary-classification` task ID to distinguish from multiclass and multilabel tasks.

### DIFF
--- a/js/src/lib/interfaces/Types.ts
+++ b/js/src/lib/interfaces/Types.ts
@@ -102,6 +102,10 @@ export const PIPELINE_DATA = ensureRecordOfPipelines({
 				name: "Intent Classification",
 			},
 			{
+				type: "binary-classification",
+				name: "Binary Classification",
+			},
+			{
 				type: "multi-class-classification",
 				name: "Multi Class Classification",
 			},
@@ -377,6 +381,10 @@ export const PIPELINE_DATA = ensureRecordOfPipelines({
 				type: "multi-class-image-classification",
 				name: "Multi Class Image Classification",
 			},
+			{
+				type: "binary-image-classification",
+				name: "Binary Image Classification",
+			},
 		],
 		modality: "cv",
 		color:    "blue",
@@ -458,6 +466,10 @@ export const PIPELINE_DATA = ensureRecordOfPipelines({
 			{
 				type: "tabular-multi-label-classification",
 				name: "Tabular Multi Label Classification",
+			},
+			{
+				type: "tabular-binary-classification",
+				name: "Tabular Binary Classification",
 			},
 		],
 		color: "blue",


js/src/lib/interfaces/Types.ts
```diff
@@ -377,6 +381,10 @@ export const PIPELINE_DATA = ensureRecordOfPipelines({
 				type: "multi-class-image-classification",
 				name: "Multi Class Image Classification",
 			},
+			{
+				type: "binary-image-classification",
```
lhoestq, line 385: (just curious) If I'm not mistaken this is a supported task in autotrain right ? (according to https://github.com/huggingface/autonlp-backend/blob/49bd7f159120ce8c11b0af0cf55083469ff21d22/zeus/scripts/create_evaluation_project.py#L20-L33)

js/src/lib/interfaces/Types.ts
```diff
@@ -459,6 +467,10 @@ export const PIPELINE_DATA = ensureRecordOfPipelines({
 				type: "tabular-multi-label-classification",
 				name: "Tabular Multi Label Classification",
 			},
+			{
+				type: "tabular-binary-classification",
```
lhoestq, line 471: This doesn't seem to be a task in autotrain ? it makes sense to have it here though anyway :)

lhoestq (approved):
Thanks ! I'll let @abhishekkrthakur and @SBrandeis say if it matches the taxonomy they're aiming for in autotrain

js/src/lib/interfaces/Types.ts
lewtun: It's soon to be released :) 

In AutoTrain it's called `image_binary_classification`, but we should align it with the `datasets` / Hub taxonomy IMO: https://github.com/huggingface/autonlp-backend/blob/918474d3e0ee914f7ebc50f6d89e20397eacd95e/database/database/enums.py#L126

js/src/lib/interfaces/Types.ts
lewtun: Yes, it is defined here: https://github.com/huggingface/autonlp-backend/blob/918474d3e0ee914f7ebc50f6d89e20397eacd95e/database/database/enums.py#L120

js/src/lib/interfaces/Types.ts
lhoestq: If it's just a matter of having dashes in YAML and underscores in python it's fine I think ;)

lewtun: Yes, it's mostly about reordering the words :)

osanseviero:
Closing this PR as it's been outdated. This part of the code now lives in https://github.com/huggingface/hub-docs. Thanks for the contribution!

PR #227, opened 2022-07-07
integration of pythae to hf-hub

clementchadebec:
As discussed with @osanseviero,  I am opening this PR as part of the integration process of the [pythae library](https://github.com/clementchadebec/benchmark_VAE) to the hub (see https://github.com/clementchadebec/benchmark_VAE/issues/24)

### DIFF
--- a/js/src/lib/interfaces/Libraries.ts
+++ b/js/src/lib/interfaces/Libraries.ts
@@ -26,6 +26,7 @@ export enum ModelLibrary {
 	"fasttext"               = "fastText",
 	"stable-baselines3"      = "Stable-Baselines3",
 	"ml-agents"              = "ML-Agents",
+	"pythae"				 = "pythae"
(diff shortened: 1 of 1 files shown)


osanseviero (approved):
Very cool! :fire:

osanseviero:
@clementchadebec how many existing models are there on the Hub with the tag at the moment?

osanseviero:
Cool! I'm ok with merging and deploying this PR, but I'm a bit hesitant of deploying something that will add a filter button that leads to a single model. Would you be ok if I wait until there are a couple more?

NimaBoscarino:
Looks like there are [some more models now](https://huggingface.co/models?other=pythae)! @osanseviero do we think it's ready to deploy?

osanseviero:
Yes let's merge it!

PR #230, opened 2022-07-08
Remove outdated section on programmatic access

BenjaminBossan:
This section makes it seem like programmatic access is still not
available, when in fact it is. Instead of updating the section, I opted
for removing it because the same topic is already covered under "next
steps":

https://huggingface.co/docs/hub/repositories-next-steps#how-to-programmatically-manage-repositories


julien-c (commented):
this looks good to me, but note that in Next Steps i feel like Programmatic access should be at the top – most important item (and link to correct doc pages for `huggingface_hub`)

julien-c:
@BenjaminBossan oh yeah that doc isn't merged yet but it's the one in https://github.com/huggingface/huggingface_hub/pull/930

julien-c (approved):
i think this could/should be merged now, right?

BenjaminBossan:
Ah yes, I forgot about this. One of the CI checks is failing, but I don't understand how it relates to the changes.

julien-c:
CI fail is certainly unrelated, feel free to merge! Docs will be auto-updated in hf.co

BenjaminBossan:
Done

PR #231, opened 2022-07-08
some more `master`=>`main` switch

julien-c:
following https://github.com/huggingface/datasets/pull/4643 but also did some forgotten ones from before


osanseviero (approved):
Thanks for the changes! Note that the UI still says master, and main redirects to master https://huggingface.co/docs/datasets/master/en/index

julien-c:
yep @osanseviero that should be solved by a PR that was just merged in moon-landing

PR #250, opened 2022-07-19
Inference widget was using a value that was one keypress behind the current state

ehrencrona:
The inference widget was submitting a value that was from one keypress before the last one. 

To reproduce:

* go to https://huggingface.co/gpt2 and type the text "123" at the end of the example text in the inference widget
* Submit
* Check the URL: you will see that it only submitted "12". 

This is because of how content editable events work; listening to the `input` event rather than the keypress fixes it.

I've tested this on Chrome and Safari on Mac but no other browsers.


julien-c:
cc @mishig25 for when he's back! (also cc @gary149 for visibility)

PR #251, opened 2022-07-20
Fix `WidgetTextArea`

mishig25:
Fixes 2 bugs found on `TextWidgetArea`:
1. `text content` of `contenteditable` span was excluding last char. for example: should be `abc`, was `ab`
2. `text content` of `contenteditable` span was wrong after: "once I have received the previous output, I deleted what has been generated but it was not taken out of what that have been sent to the API"

Luckily found a fix on https://stackoverflow.com/a/51702629:
`whitespace-pre-wrap inline-block` tailwind classes are needed to get correct newlines from `el.textContent` on Chrome where `el.contentEditable == true`, which eliminated the need to use custom `getTextContent` from [here](https://stephenhaney.com/2020/get-contenteditable-plaintext-with-correct-linebreaks/), which was causing the second bug mentioned above

### DIFF
--- a/js/src/lib/components/InferenceWidget/shared/WidgetTextarea/WidgetTextarea.svelte
+++ b/js/src/lib/components/InferenceWidget/shared/WidgetTextarea/WidgetTextarea.svelte
@@ -9,8 +9,6 @@
 	export let isLoading = false;
 	export let size: "small" | "big" = "small";
 
-	let newValue = "";
-	let isOnFreshLine = true;
 	let containerSpanEl: HTMLSpanElement;
 	const typingEffectSpeedMs = 12;
 	const classNamesInput = "font-normal text-black dark:text-white";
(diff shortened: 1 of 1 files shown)


js/src/lib/components/InferenceWidget/shared/WidgetTextarea/WidgetTextarea.svelte
```diff
@@ -110,20 +73,21 @@
 
 <WidgetLabel {label}>
 	<svelte:fragment slot="after">
+		<!-- `whitespace-pre-wrap inline-block` are needed to get correct newlines from `el.textContent` on Chrome -->
 		<span
 			class="{isLoading ? 'pointer-events-none' : ''} {label
 				? 'mt-1.5'
 				: ''} block overflow-auto resize-y py-2 px-3 w-full {size === 'small'
 				? 'min-h-[42px]'
-				: 'min-h-[144px]'} max-h-[500px] whitespace-pre-wrap border border-gray-200 rounded-lg shadow-inner outline-none focus:ring focus:ring-blue-200 focus:shadow-inner dark:bg-gray-925"
```
mishig25, line 118: added this accidental omission of `border` back in https://github.com/huggingface/hub-docs/commit/d2c0de490c2be461569a3393c0f93c030f5c51bf

PR #263, opened 2022-07-27
Implement VisualQA widget

mishig25:
### Implement VisualQuestionAnswering widget

demo: fist widget [here](https://62e138f1056cc019e97ad654--huggingface-widgets.netlify.app/)

input (similar to [QuestionAnsweringWidget](https://github.com/huggingface/hub-docs/blob/5686c85b0ddbb8e4b95d9718b4b2c6d626aa46c8/js/src/lib/components/InferenceWidget/widgets/QuestionAnsweringWidget/QuestionAnsweringWidget.svelte#L81-L83) instead of context , there is an image (base64 str) ):
```js
const requestBody = {
    inputs: { 
        question: string, 
        image: string, // base64 str of img 
    },
};
```

output1 (similar to [ImageClassificationWidget](https://github.com/huggingface/hub-docs/blob/d13a2a5bcf50a6431b74f972e1806c0fffccb5f6/js/src/lib/components/InferenceWidget/widgets/ImageClassificationWidget/ImageClassificationWidget.svelte#L80)):
```js
Array<{ label: string; score: number }>
```

Screenshot 2022-07-27 at 15 02 20

todos:
- [x] test when api-inference vqa is up
- [x] test widget input samples
- [ ] document widget input sample

cc: @philschmid @NielsRogge

### DIFF
--- a/docs/hub/models-widgets-examples.md
+++ b/docs/hub/models-widgets-examples.md
@@ -248,6 +248,16 @@ widget:
   example_title: "Dog"
 ```
 
+### Visual Question Answering
+
+```yaml
+widget:
+- text: "What animal is it?"
+  src: "https://huggingface.co/datasets/mishig/sample_images/resolve/main/tiger.jpg"
+- text: "Where is it?"
+  src: "https://huggingface.co/datasets/mishig/sample_images/resolve/main/palace.jpg"
+```
+
 ## Other
 
 ### Structured Data Classification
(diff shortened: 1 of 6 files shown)


julien-c (commented):
I'm very late to the party, but is this the right input/output interface for a general version of `visual-question-answering`? 

vs. something that would take `[image, string]` and return `string`, like question-answering and text-generation?

(i guess this question would have been more relevant on the corresponding pipeline PR...)

also cc @osanseviero

julien-c:
Pipeline PR for reference: https://github.com/huggingface/transformers/pull/17286

Narsil:
> I'm very late to the party, but is this the right input/output interface for a general version of visual-question-answering?

Ofc it's always debattable.

But I think that here it makes sense.

`text-generation` actually outputs several strings when using `num_return_sequences` (which is nice in an autocomplete settings) (And having outputs depending on there parameters/inputs is kind of tricking when using the pipelines, since the caller has to unwrap the results differently based on the parameters)

`question-answering` is great when it works, but when it doesn't it's kind of hard to do anything about it. Using `scores` + `top_k` enables `visual-question-answering` to actually tell multiple things for instance:

Screenshot from 2022-08-02 19-09-21

Here both flowers and fountain are actually correct. 

Another example:
Screenshot from 2022-08-02 19-11-08

PR #269, opened 2022-08-02
ENH Updated resources

merveenoyan:
Updated resources sections of various tasks.

### DIFF
--- a/tasks/src/image-classification/about.md
+++ b/tasks/src/image-classification/about.md
@@ -25,6 +25,7 @@ clf("path_to_a_cat_image")
 ## Useful Resources
 
 - [Fine-Tune ViT for Image Classification with 🤗Transformers](https://huggingface.co/blog/fine-tune-vit)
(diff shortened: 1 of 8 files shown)


tasks/src/image-classification/about.md
```diff
@@ -25,6 +25,7 @@ clf("path_to_a_cat_image")
 ## Useful Resources
 
 - [Fine-Tune ViT for Image Classification with 🤗Transformers](https://huggingface.co/blog/fine-tune-vit)
+- [Computer Vision in Hugging Face - CV Study Group](https://www.youtube.com/watch?v=oL-xmufhZM8)
```
osanseviero, line 28: Nice! Maybe that video should be renamed, not sure though!

osanseviero (approved):
Thanks for this!

osanseviero (approved):
LG, thanks! I left some minor points Btw we don't have the convention of adding ENH in our codebase (not in hub-docs at least)

PR #273, opened 2022-08-03
Add `tags` to the Spaces configuration reference

pcuenca:
I kept it simple rather than explaining that users can manually filter using a URL param.


osanseviero (approved):
Looks good to me! Thanks! :fire:

julien-c:
Cool addition, thanks!

PR #274, opened 2022-08-03
Add docs for library_name

### LINKED ISSUE #264: Add docs about library name
cc @osanseviero @philschmid @julien-c 

There is no `library_name` mentioned in https://huggingface.co/docs/hub/models-cards

The library name is added to the tags of the model on the hub. It's determined as such:

1. The user can force `library_name` in the model card
2. If one of the tags in the model card already is a library name, then do nothing
3. We try to autodetect based on the presence of files for diffusers, keras, sklearn and nemo
4. Lastly, if nothing was detected and there is a valid `config.json` file, the library name is set to `"transformers"`

osanseviero:
Fixes #264

### DIFF
--- a/docs/hub/models-cards.md
+++ b/docs/hub/models-cards.md
@@ -35,23 +35,41 @@ metrics:
 ---
 ```
 
-You can also specify the supported frameworks in the model card metadata section:
+The metadata that you add to the model card enables certain interactions on the Hub. For example:
+* Allow users to filter and discover models at https://huggingface.co/models.
+* If you choose a license using the keywords listed in the right column of [this table](./repositories-licenses), the license will be displayed on the model page.
+* Adding datasets to the metadata will add a message reading `Datasets used to train:` to your model card and link the relevant datasets, if they're available on the Hub.
+
+Dataset, metric, and language identifiers are those listed on the [Datasets](https://huggingface.co/datasets), [Metrics](https://huggingface.co/metrics) and [Languages](https://huggingface.co/languages) pages and in the [`datasets`](https://github.com/huggingface/datasets) repository.
+
+See the detailed model card specification [here](https://github.com/huggingface/hub-docs/blob/main/modelcard.md?plain=1).
+
+### Specifying a library
+
+You can also specify the supported libraries in the model card metadata section. Find more about our supported libraries [here](./models-libraries). The library can be specified with the following order of priority
+
+1. Specifying `library_name` in the model card (recommended)
+
+```yaml
+library_name: flair
+```
+
+2. Having a tag with the name of a library that is supported
 
 ```yaml
 tags:
 - flair
 ```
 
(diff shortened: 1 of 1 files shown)


docs/hub/models-cards.md
```diff
@@ -35,23 +35,41 @@ metrics:
...
+
+2. Having a tag with the name of a library that is supported
 
 ```yaml
 tags:
 - flair
 ```
 
-Find more about our supported libraries [here](./models-libraries), and see the detailed model card specification [here](https://github.com/huggingface/hub-docs/blame/main/modelcard.md).
+If it's not specified, the Hub will try to automatically detect the library type. This approach is discouraged and repo creators should use the explicit license as much as possible.
```
coyotte508, line 64:
```suggestion
If it's not specified, the Hub will try to automatically detect the library type. This approach is discouraged and repo creators should use the explicit `library_name` as much as possible.
```

coyotte508 (approved):

docs/hub/models-cards.md
```diff
@@ -35,23 +35,41 @@ metrics:
...
+
+Dataset, metric, and language identifiers are those listed on the [Datasets](https://huggingface.co/datasets), [Metrics](https://huggingface.co/metrics) and [Languages](https://huggingface.co/languages) pages and in the [`datasets`](https://github.com/huggingface/datasets) repository.
+
+See the detailed model card specification [here](https://github.com/huggingface/hub-docs/blob/main/modelcard.md?plain=1).
+
+### Specifying a library
+
+You can also specify the supported libraries in the model card metadata section. Find more about our supported libraries [here](./models-libraries). The library can be specified with the following order of priority
+
+1. Specifying `library_name` in the model card (recommended)
```
julien-c, line 51:
```suggestion
1. Specifying `library_name` in the model card (recommended if your model is not a `transformers` model)
```

docs/hub/models-cards.md
```diff
@@ -35,23 +35,41 @@ metrics:
...
+
+2. Having a tag with the name of a library that is supported
 
 ```yaml
 tags:
 - flair
 ```
 
-Find more about our supported libraries [here](./models-libraries), and see the detailed model card specification [here](https://github.com/huggingface/hub-docs/blame/main/modelcard.md).
+If it's not specified, the Hub will try to automatically detect the library type. This approach is discouraged and repo creators should use the explicit `library_name` as much as possible.
```
julien-c, line 64:
```suggestion
If it's not specified, the Hub will try to automatically detect the library type. Unless your model is from `transformers`, this approach is discouraged and repo creators should use the explicit `library_name` as much as possible.
```

julien-c (commented):
I don't necessarily agree with the recommendation to always specify `library_name`, notably because 80k out of 120k models are from `transformers` and they won't have a `library_name` in the short/medium term. 

So the current coverage of this metadata item is bound to be low.

As a middleground, I propose the following suggestions. Let me know if this would work for you! 🤟

osanseviero:
Yes, this works for me! Merged your changes

julien-c (approved):

PR #282, opened 2022-08-15
Add keras to Moldel integration table

philschmid:


osanseviero (changes requested):
Thanks for this PR! The API and the widgets don't work great out of the box for Keras models at the moment, so I would remove it.

osanseviero (approved):
Thanks a lot for this PR!

PR #290, opened 2022-09-02
Implement img2text widget

mishig25:
### Implement Text-to-Image widget

https://github.com/huggingface/transformers/pull/18821#issuecomment-1234524309

input (identical to [ImageClassificationWidget](https://github.com/huggingface/hub-docs/blob/d13a2a5bcf50a6431b74f972e1806c0fffccb5f6/js/src/lib/components/InferenceWidget/widgets/ImageClassificationWidget/ImageClassificationWidget.svelte#L45)):
```js
const requestBody = { file }; // img file
```

output (identical to [TextGenerationWidget](https://github.com/huggingface/hub-docs/blob/21965aab98c78c5b6971371bde0b8f17853fab40/js/src/lib/components/InferenceWidget/widgets/TextGenerationWidget/TextGenerationWidget.svelte#L158)):
```js
Array<{ generated_text: string; }>
```
example:
```js
[
  {
    "generated_text": "Some text"
  }
]
```

### Note:

In transformers implementation, I see that the pipeline is called [image-to-text-generation](https://github.com/OlivierDehaene/transformers/blob/dc0cc5322f55901265f816b5eda45f73f77d473b/src/transformers/pipelines/image2text_generation.py#L29), while in the hub, we already defined it as [image-to-text](https://huggingface.co/models?pipeline_tag=image-to-text) without the **generation** part. Please let me know if it is an issue: @Narsil @osanseviero 

todos:
- [ ] test when api-inference is up @Narsil 
- [ ] document widget input sample

### DIFF
--- a/js/src/lib/components/InferenceWidget/InferenceWidget.svelte
+++ b/js/src/lib/components/InferenceWidget/InferenceWidget.svelte
@@ -10,6 +10,7 @@
 	import FeatureExtractionWidget from "./widgets/FeatureExtractionWidget/FeatureExtractionWidget.svelte";
 	import FillMaskWidget from "./widgets/FillMaskWidget/FillMaskWidget.svelte";
 	import ImageClassificationWidget from "./widgets/ImageClassificationWidget/ImageClassificationWidget.svelte";
+	import ImageToTextWidget from "./widgets/ImageToTextWidget/ImageToTextWidget.svelte";
 	import ImageSegmentationWidget from "./widgets/ImageSegmentationWidget/ImageSegmentationWidget.svelte";
 	import ObjectDetectionWidget from "./widgets/ObjectDetectionWidget/ObjectDetectionWidget.svelte";
 	import QuestionAnsweringWidget from "./widgets/QuestionAnsweringWidget/QuestionAnsweringWidget.svelte";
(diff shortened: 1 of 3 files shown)


osanseviero:
Related to

> In transformers implementation, I see that the pipeline is called [image-to-text-generation](https://github.com/OlivierDehaene/transformers/blob/dc0cc5322f55901265f816b5eda45f73f77d473b/src/transformers/pipelines/image2text_generation.py#L29), while in the hub, we already defined it as [image-to-text](https://huggingface.co/models?pipeline_tag=image-to-text) without the generation part. Please let me know if it is an issue: 

This should take care of it https://github.com/huggingface/transformers/pull/18864

Narsil:
LGTM. I don't think we chose `List[{"generated_text": str}]` for the output  but `{"generated_text": str}` only (I think).

I think being aligned with what we already have is better.

@OlivierDehaene In general, unfortunately since the pipelines were created at different times by different people originally, the exact output types are not super normalized, especially on the List, List of List stuff. This is something I would like to change on v5 (whenever we decide to go for it and I don't think we have plans) so that the pipeline code could be more regular.

For `text-generation` it can really generate multiple texts (it's controlled by `num_return_sequences`) and I think enabling the list here also enables more uses cases (even for captioning you might be interested to generate several at once and choose the best). 

However, I don't think we should by highly strict in v4 since every pipeline is ever so slightly different than it's neighbor.

osanseviero:
Closing as this now lives in https://github.com/huggingface/hub-docs.

PR #292, opened 2022-09-02
Add SKOPS as library

Kakulukian:
Adding skops in the library filter on the hub

cc @osanseviero @adrinjalali

### DIFF
--- a/js/src/lib/interfaces/Libraries.ts
+++ b/js/src/lib/interfaces/Libraries.ts
@@ -16,6 +16,7 @@ export enum ModelLibrary {
 	"pyannote-audio"         = "pyannote.audio",
 	"sentence-transformers"  = "Sentence Transformers",
 	"sklearn"                = "Scikit-learn",
+	"skops"                  = "SKOPS",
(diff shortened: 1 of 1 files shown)


js/src/lib/interfaces/Libraries.ts
```diff
@@ -16,6 +16,7 @@ export enum ModelLibrary {
 	"pyannote-audio"         = "pyannote.audio",
 	"sentence-transformers"  = "Sentence Transformers",
 	"sklearn"                = "Scikit-learn",
+	"skops"                  = "SKOPS",
```
adrinjalali, line 19: I think we've written `skops` all small letters everywhere. Otherwise, I'm not sure if we want to add it here: https://github.com/huggingface/moon-landing/issues/3757

Kakulukian: I opened the pr before the discussion, I will close it ;)

julien-c:
i think this could be reopened as i think in the attached issue we said we could add it to the filter in /models when we want to

PR #293, opened 2022-09-05
Widgets call inference run `onMount`

mishig25:
TLDR: call inference run on widgets using `callApiOnMount` prop & flag [X-Load-Model = 0](https://github.com/huggingface/hub-docs/blob/4cb0e4dae893d905bf7a3a780afef231f897ce6f/js/src/lib/components/InferenceWidget/shared/helpers.ts#L83-L85)

`callApiOnMount` prop already exists on most widgets like [TextGenerationWidget](https://github.com/huggingface/hub-docs/blob/21965aab98c78c5b6971371bde0b8f17853fab40/js/src/lib/components/InferenceWidget/widgets/TextGenerationWidget/TextGenerationWidget.svelte) & this PR adds `callApiOnMount ` prop to remaining widgets like [ImageSegmentationWidget](https://github.com/huggingface/hub-docs/blob/7bda9b6965e26a57dbd9cee805577d6f0c84feee/js/src/lib/components/InferenceWidget/widgets/ImageSegmentationWidget/ImageSegmentationWidget.svelte)

when set to `true`, `callApiOnMount` makes the widget load example input & call the api. And since commonly used models have cached output, users should receive output instantly

### DIFF
--- a/js/src/lib/components/InferenceWidget/widgets/ImageClassificationWidget/ImageClassificationWidget.svelte
+++ b/js/src/lib/components/InferenceWidget/widgets/ImageClassificationWidget/ImageClassificationWidget.svelte
@@ -36,7 +36,11 @@
 		getOutput(file);
 	}
 
-	async function getOutput(file: File | Blob, withModelLoading = false) {
+	async function getOutput(
+		file: File | Blob,
+		withModelLoading = false,
+		isOnLoadCall = false
+	) {
 		if (!file) {
 			return;
 		}
@@ -59,7 +63,8 @@
 			apiToken,
 			parseOutput,
 			withModelLoading,
-			includeCredentials
+			includeCredentials,
+			isOnLoadCall
 		);
 
 		isLoading = false;
@@ -119,7 +124,7 @@
 			if (callApiOnMount && src) {
 				imgSrc = src;
 				const blob = await getBlobFromUrl(imgSrc);
-				getOutput(blob);
+				getOutput(blob, false, true);
 			}
 		})();
 	});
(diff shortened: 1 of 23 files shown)


js/src/lib/components/InferenceWidget/widgets/ImageSegmentationWidget/ImageSegmentationWidget.svelte
```diff
@@ -223,10 +228,17 @@
 		outputJson = "";
 	}
 
-	onMount(() => {
+	onMount(async () => {
```
Pierrci, line 231: I think this is a pattern that should be avoided if possible? cc @pngwn @beurkinger 

(particularly when it returns something to be called on destroy, though this is not the case here)

mishig25: thanks for catching that! https://github.com/huggingface/hub-docs/pull/293/commits/de0396992e759e0f75eb617c5f1ef14682f0e56f fixes those instances

js/src/lib/components/InferenceWidget/widgets/ImageClassificationWidget/ImageClassificationWidget.svelte
```diff
@@ -106,6 +112,17 @@
...
+
+	onMount(() => {
+		async () => {
+			const [src] = getDemoInputs(model, ["src"]);
+			if (callApiOnMount && src) {
+				imgSrc = src;
+				const blob = await getBlobFromUrl(imgSrc);
+				getOutput(blob);
+			}
+		};
```
Pierrci, line 124: I think it needs to be called no? Here it's just defined?

mishig25: oh indeed, handled in 54cf71564356fd98a7fad36121d13029db276b29
thanks for catching it!

PR #295, opened 2022-09-05
Fix link to pickletools.genops in pickle scanning docs

mmaitre314:
Make the link to the `pickletools.genops` docs clickable.


McPatate (approved):
Thanks a lot !

PR #301, opened 2022-09-12
remove `SavedModel` as it also serializes code no?

julien-c:


julien-c:
or even better mention that `SavedModel` is not an option as it also serializes user code:

image

McPatate:
> or even better mention that `SavedModel` is not an option as it also serializes user code:


Ah yes nice catch !

julien-c:
Closing as this was done by @McPatate in #334

PR #278, opened 2022-08-09
Added files under each task page and created tabular-classification task page

merveenoyan:
FYI @julien-c @osanseviero

PR #302, opened 2022-09-12
Tabular Classification task page

merveenoyan:
Split this PR into two: https://github.com/huggingface/hub-docs/pull/278


merveenoyan:
We only need the part for input-output @gary149 @beurkinger

tasks/src/tabular-classification/about.md
```diff
@@ -0,0 +1,32 @@
...
+
+### Fraud Detection
+Tabular classification models can be used in detecting fraudulent credit card transactions, where features can be amount of the transaction and account balance, and target to predict is whether the transaction is fraudulent or not. 
+
+### Churn Prediction
+Tabular classification models can be used in predicting customer churn in telecommunication. An example dataset for the task is hosted [here](https://huggingface.co/datasets/scikit-learn/churn-prediction).
+
+## Model Hosting and Inference
+
+You can use the [Skops](https://github.com/skops-dev/skops) library to host `scikit-learn` models on Hugging Face Hub. With `Skops`, you can automatically generate a configuration for your model that includes the information needed to enable the widget on Hugging Face Hub and the requirements needed to infer the model. `Skops` also generates a model card including a hyperparameter table and plot of the model, that you can later add information programmatically. You can push the model and the related files using Skops. The repository contains the configuration file, model card and the model. You can pull a `scikit-learn` `tabular-classification` model like below:
```
osanseviero, line 11: I would suggest to read this with the eyes of the intended audience, as people might not know many of the words here (what's a model card? what's a hyperparamerter? etc) and tune it accordingly.

tasks/src/tabular-classification/data.ts
```diff
@@ -0,0 +1,54 @@
...
+			id:          "scikit-learn/adult-census-income",
+		},
+		{
+			description: "Multi-class dataset on iris flower species.",
+			id:          "scikit-learn/iris",
+		},
+	],
+	demo: {
+		inputs: [],
+		outputs: [],
```
osanseviero, line 19: You can already add an output as done for other classification tasks

tasks/src/tabular-classification/data.ts
```diff
@@ -0,0 +1,54 @@
...
+		},
+
+	],
+	models: [
+		{
+			description: "Breast cancer prediction model based on decision trees.",
+			id:          "scikit-learn/cancer-prediction-trees",
+		},
+	],
+	summary:      "Tabular classification is the task of classifying a target category based on a set of numerical or categorical features.",
```
osanseviero, line 49: Let's follow up the discussion about the summary from the other PR here. We were discussing categorical and numerical features might sound confusing to the expected audience

osanseviero (commented):
Thanks for working on this! I left some suggestions to improve this :)

tasks/src/tabular-classification/data.ts
merveenoyan: but I don't know how the dataframe component should look like, since we didn't have it previously. pinging @gary149

tasks/src/tabular-classification/data.ts
merveenoyan: I addressed this by adding more context 🙂

tasks/src/tabular-classification/about.md
```diff
@@ -0,0 +1,41 @@
+## About the Task
+
+Tabular classification is the task of predicting a variable that can take limited number of values. This type of variable is called a categorical variable. Tabular classification models take a table of data as input and output a category. For example, input can be attributes of a customer (balance of the customer, their education status, or more) and output can be whether customer churns or not.
+There's three types of categorical variables:
+
+- Binary variables: Variables that can take two values, like yes or no, open or closed. The task of predicting binary variables is called binary classification.
+- Ordinal variables: Variables that have a ranking relationship among them, e.g. good, insignificant and bad product reviews. The task of predicting ordinal variables is called ordinal classification.
+- Nominal variables: Variables that have no ranking relationship among them, e.g. predicting an animal from their weight and height, where categories are cat, dog or bird. The task of predicting nominal variables is called multinomial classification.
```
osanseviero, line 8: Is knowing the different types of categorical variables very important for the audience?

tasks/src/tabular-classification/about.md
```diff
@@ -0,0 +1,41 @@
...
+
+## Use Cases
+
+### Fraud Detection
+Tabular classification models can be used in detecting fraudulent credit card transactions, where features can be amount of the transaction and account balance, and target to predict is whether the transaction is fraudulent or not. This is an example to binary classification.
```
osanseviero, line 13:
```suggestion
Tabular classification models can be used in detecting fraudulent credit card transactions, where the features could be the amount of the transaction and the account balance, and the target to predict could be whether the transaction is fraudulent or not. This is an example of binary classification.
```

tasks/src/tabular-classification/data.ts
osanseviero: Alright! I was thinking we could pass just the items of a single row if that helped keep things simpler though. 

E.g. for text classification you could also pass a batch of samples, but the example we show it's just a single example. Also in inference in production, many times the model will just infer from one sample at a time.

osanseviero (commented):
Thanks for the changes! Some suggestions :)

tasks/src/tabular-classification/about.md
merveenoyan: yes it determines the problem you're solving (see every last sentence) I think it's good to know and not that advanced 😅

tasks/src/tabular-classification/data.ts
```diff
@@ -0,0 +1,54 @@
...
+			description: "Breast cancer prediction model based on decision trees.",
+			id:          "scikit-learn/cancer-prediction-trees",
+		},
+	],
+	summary:      "Tabular classification is the task of classifying a target category (a group) based on set of attributes.",
+	widgetModels: [""],
+    youtubeId:""
+};
+
+export default taskData;
```
beurkinger, line 54: missing EOF

beurkinger:
Can you put your changes in a branch, instead of cloning the whole repo, please ? It would make reviewing and working on your changes much easier. Thank you !

PR #318, opened 2022-09-20
add 8080 port in the supported ports for spaces

christophe-rannou:
Add to documentation that port 8080 is now opened by default for spaces


osanseviero (approved):
Thanks!

PR #321, opened 2022-09-21
Rename the models-cards and models-cards-co2 pages and adjust all links accordingly.

NimaBoscarino:
For UX purposes, the link should read "model-cards" since that's what these are known as, and having it read "models-cards" is confusing. Likewise for models-cards-co2.

A PR with a redirect route has been created on the backend, to handle any third-party linking (or linking from other HF documentation/repos).


julien-c (commented):
lgtm

nateraw (approved):
LGTM 🚀

Thanks

PR #326, opened 2022-09-23
[doc+snippets] No more `use_auth_token=True` needed

julien-c:
Keeping as draft until https://github.com/huggingface/huggingface_hub/pull/1064 lands in a release. (and subsequently in a transformers or diffusers release if those still pin a version of `huggingface_hub` – not sure right now)

### DIFF
--- a/docs/hub/models-gated.md
+++ b/docs/hub/models-gated.md
@@ -15,7 +15,9 @@ from huggingface_hub import notebook_login
 notebook_login()
 ```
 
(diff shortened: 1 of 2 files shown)


julien-c:
N.B.: This is for simplicity, passing `use_auth_token` is stil supported (with same potential params as before), it's just that it now defaults to True

osanseviero (approved):
LG! :fire:

patrickvonplaten:
Nice! @patil-suraj @pcuenca @anton-l let's make sure we update the Diffusers docs here as well

julien-c:
Making note to merge this around end-of-year!

PR #334, opened 2022-09-28
fix(hub): remove h5 and savedmodels from safe format list

McPatate:
Thanks to @mmaitre314 for spotting that H5/SavedModels is not a safe serialisation format, as specified on [this documentation](https://github.com/tensorflow/tensorflow/blob/master/SECURITY.md#running-untrusted-models).

### DIFF
--- a/docs/hub/security-pickle.md
+++ b/docs/hub/security-pickle.md
@@ -208,8 +208,6 @@ model = AutoModel.from_pretrained("bert-base-cased", from_flax=True)
 
 ### Use your own serialization format
 
-- [HDF5](https://github.com/HDFGroup/hdf5)
-- [SavedModel](https://www.tensorflow.org/guide/saved_model)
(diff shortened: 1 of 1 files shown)


merveenoyan:
The good thing about SavedModel is that when you have custom code (or custom layer defined inbetween) if you haven't passed that during deserialization (sort of acknowledge), you'll get a warning. (from what I remember and also docs: https://www.tensorflow.org/guide/keras/save_and_serialize) That's what makes it safer. Every model includes code and when you serialize it you'll eventually get custom code deserialized so technically every format should be unsafe. The serialization technique skops and recently pytorch came up with works in same principle too. Also pinging @julien-c

merveenoyan:
@mmaitre314 ah I took a look at the docs, it also encourages to subclass for this reason I guess.

julien-c (approved):

PR #344, opened 2022-09-30
feat(security): add disclaimer about imports safety

McPatate:
Adding a disclaimer as discussed in https://github.com/huggingface/moon-landing/pull/3814.


osanseviero (approved):
Thanks!

PR #348, opened 2022-10-03
Add docs about the /commit endpoint

coyotte508:
cc @Wauplin 

Any reason for the trailing `   	|` for each line in the table?


julien-c:
include a screenshot :)

coyotte508:
The docs are here: https://moon-ci-docs.huggingface.co/docs/hub/pr_348/en/api

julien-c:
i would refer to a detailed section below the table because this is not super legible:

image

coyotte508:
or we could overhaul the docs and use a swagger-like interface x)

NimaBoscarino:
I don't think there's any reason for the trailing `|`, as far as I know. Overhauling the endpoints docs to use something like Swagger would be awesome 👍 In the meantime though I agree that the payload is too long for the table, so maybe defining the `json` portion under the table would be cool, like Julien suggested.

osanseviero:
Closing this as it's outdated/stalled

PR #376, opened 2022-10-06
First commit of task: unconditional-image-generation.

juancopi81:
Hi,

This is my first pull request. I am working on the docs for the unconditional image generator. I added the folders (in src and assets) and the data.ts file


tasks/src/unconditional-image-generation/data.ts
```diff
@@ -0,0 +1,55 @@
+import type { TaskDataCustom } from "../Types";
+
+const taskData: TaskDataCustom = {
+	datasets: [
+		{
+			description: "NFT images dataset of 'cryptopunks' for unconditional image generation.",
```
merveenoyan, line 6: It would be nice to have a generic dataset like CIFAR or CelebA. https://huggingface.co/datasets/cifar100 this is more of a general purpose dataset that would be useful for use cases 🙂

tasks/src/unconditional-image-generation/data.ts
```diff
@@ -0,0 +1,55 @@
...
+		{
+			description: "NFT images dataset of 'alpacadabraz' for unconditional image generation.",
+			id:          "huggingnft/alpacadabraz",
+		},
+	],
+	demo: {
+		inputs: [
+			{
+				label:   "Input",
+				content: "",
```
merveenoyan, line 18: You can check out this demo for inspiration on how it should look like 🙂 https://huggingface.co/spaces/fusing/celeba-diffusion

tasks/src/unconditional-image-generation/data.ts
```diff
@@ -0,0 +1,55 @@
...
+			description: "LightWeight GAN model for unconditional generation.",
+			id:          "huggingnft/cryptopunks",
+		},
+		{
+			description: "The mode generates high-quality image synthesis using diffusion probabilistic models, a class of latent variable models inspired by considerations from nonequilibrium thermodynamics.",
+			id:          "google/ddpm-celebahq-256",
+		},
+	],
+	summary:      "Unconditional image generation is the task of generating images with no condition in any context (prompt text, another image, etc.). Once trained, the model will create images that resemble its training data distribution.",
+	widgetModels: ["google/vit-base-patch16-224"],
```
merveenoyan, line 50: I think this model is an `image-classification` model. We look for unconditional-image-generation models. However, there's no widget for them as of now, so it's okay to leave it as is 🙂
```suggestion
	widgetModels: "",
```

tasks/src/unconditional-image-generation/data.ts
```diff
@@ -0,0 +1,55 @@
...
+			id:          "Inception score (IS)",
+		},
+		{
+			description: "The Fréchet Inception Distance (FID) evaluates the quality of images created by a generative model by calculating the distance between feature vectors for real and generated images.",
+			id:          "Frećhet Inception Distance (FID)",
+		},
+	],
+	models: [
+		{
+			description: "LightWeight GAN model for unconditional generation.",
```
merveenoyan, line 41: I found these: https://huggingface.co/google/ddpm-celebahq-256 https://huggingface.co/google/ddpm-cifar10-32 especially models trained datasets like CIFAR are quite off the shelf and useful. What do you think? 🙂

tasks/src/unconditional-image-generation/data.ts
```diff
@@ -0,0 +1,55 @@
...
+		outputs: [
+			{
+				filename: "unconditional-image-generation-output.jpeg",
+				type:     "img",
+			},
+		],
+	},
+	metrics: [
+		{
+			description: "The inception score (IS) evaluates the quality of image generative models by using the output of a pre-trained classifier (InceptionV3). It measures that the generated images are diverse (the model predictions are evenly distributed across all possible labels) and distinct (the model confidently predicts a single label for each image).",
```
merveenoyan, line 31:
```suggestion
			description: "The inception score (IS) evaluates the quality of image generative models). It measures the diversity of the generated images (the model predictions are evenly distributed across all possible labels) and distinct (the model confidently predicts a single label for each image).",
```
nit 🙂

merveenoyan (commented):
Thanks a lot for working on this! 🤗 I left a few comments, can you take a look? 👀

tasks/src/unconditional-image-generation/data.ts
juancopi81: Great! Thank you very much for correction.

tasks/src/unconditional-image-generation/data.ts
juancopi81: Great! I am just noticing an extra ")" in your suggestion, so I'll change it in my PR with the other corrections if that's ok for you.

tasks/src/unconditional-image-generation/data.ts
juancopi81: Sure! I am adding CIFAR-100 and CelebA as datasets here.

tasks/src/unconditional-image-generation/data.ts
juancopi81: I agree :)! I am changing the models to these ones.

tasks/src/unconditional-image-generation/data.ts
merveenoyan: I think you can simply add a number instead. And as an output it's best if we don't use a face 🙂

juancopi81: Thanks again! Since it is better not to use a face I am now using this model https://huggingface.co/spaces/huggan/huggingnft to generate the output and also created the inputs based on it. What do you think? Thanks again for your kind support. 🙂

tasks/src/unconditional-image-generation/data.ts
```diff
@@ -0,0 +1,65 @@
...
+			description: "High-quality image generation model trained on the CIFAR-10 dataset. It synthesizes images of the ten classes presented in the dataset using diffusion probabilistic models, a class of latent variable models inspired by considerations from nonequilibrium thermodynamics.",
+			id:          "google/ddpm-cifar10-32",
+		},
+		{
+			description: "High-quality image generation model trained on the 256x256 CelebA-HQ dataset. It synthesizes images of faces using diffusion probabilistic models, a class of latent variable models inspired by considerations from nonequilibrium thermodynamics.",
+			id:          "google/ddpm-celebahq-256",
+		},
+	],
+	summary:      "Unconditional image generation is the task of generating images with no condition in any context (prompt text, another image, etc.). Once trained, the model will create images that resemble its training data distribution.",
+	widgetModels: "",
```
merveenoyan, line 60:
```suggestion
	widgetModels: [""],
```
needs to be a list 🙂

tasks/src/unconditional-image-generation/data.ts
```diff
@@ -0,0 +1,65 @@
...
+		{
+			description: "High-quality image generation model trained on the CIFAR-10 dataset. It synthesizes images of the ten classes presented in the dataset using diffusion probabilistic models, a class of latent variable models inspired by considerations from nonequilibrium thermodynamics.",
+			id:          "google/ddpm-cifar10-32",
+		},
+		{
+			description: "High-quality image generation model trained on the 256x256 CelebA-HQ dataset. It synthesizes images of faces using diffusion probabilistic models, a class of latent variable models inspired by considerations from nonequilibrium thermodynamics.",
+			id:          "google/ddpm-celebahq-256",
+		},
+	],
+	summary:      "Unconditional image generation is the task of generating images with no condition in any context (prompt text, another image, etc.). Once trained, the model will create images that resemble its training data distribution.",
```
merveenoyan, line 59:
```suggestion
	summary:      "Unconditional image generation is the task of generating images with no condition in any context (can be prompt text or another image). Once trained, the model will create images that resemble its training data distribution.",
```

tasks/src/unconditional-image-generation/data.ts
```diff
@@ -0,0 +1,65 @@
...
+			},
+			{
+				label:   "Number of rows:",
+				content: "2",
+				type:    "text",
+			},
+		],
+		outputs: [
+			{
+				filename: "unconditional-image-generation/unconditional-image-generation-output.jpeg",
```
merveenoyan, line 34: I got it up on my local and it can't find the image path of the new image. 
Ekran Resmi 2022-10-11 19 07 34
Also I think you don't necessarily need to pick model, these models usually have random seeds as inputs that later define noise to generate the image from, see this as an example: https://huggingface.co/spaces/huggan/crypto-gan 
the row parameter makes sense but maybe rename it as "number of images to generate"? Row doesn't sound clear in this context IMHO 🙂

merveenoyan (commented):
left nits 🙂

tasks/src/unconditional-image-generation/data.ts
juancopi81: Sure! Thanks for the correction :)

tasks/src/unconditional-image-generation/data.ts
juancopi81: Oh let me fix that...

juancopi81:
Thanks Merve! Hope now is better.

PR #394, opened 2022-10-11
Minor fixes on data.ts

merveenoyan:
Two commas were missing, this doesn't affect production as task page itself is not currently used (so I don't really have a screenshot) but I got it up on codespaces and error was gone.


osanseviero (approved):
Thanks!

PR #396, opened 2022-10-11
Add optimum-habana and optimum-graphcore to libraries

regisss:
Optimum Habana and Optimum Graphcore must be declared in `hub-docs/js/src/lib/interfaces/Libraries.ts` in order to enable download stats for HPUs and IPUs (see https://github.com/huggingface/moon-landing/pull/4071).

### DIFF
--- a/js/src/lib/interfaces/Libraries.ts
+++ b/js/src/lib/interfaces/Libraries.ts
@@ -26,7 +26,9 @@ export enum ModelLibrary {
 	"fasttext"               = "fastText",
 	"stable-baselines3"      = "Stable-Baselines3",
 	"ml-agents"              = "ML-Agents",
-	"pythae"                 = "Pythae"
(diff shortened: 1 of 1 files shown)


regisss:
@osanseviero @julien-c I added a `isHidden` property but I don't know where it is used. I guess it must be used somewhere to not display anything right?

julien-c:
I'm still unclear if this should be a `library` or a `filetype` so I guess i'll let @osanseviero pick one option =)

osanseviero:
As discussed internally, a `FileType` would fit better as we still expect repos to hold both this files + transformer weights

regisss:
Closing this PR as it was more convenient to use a `FileType` for `gaudi_config.json`/`ipu_config.json` than adding Optimum Habana/Graphcore to registered libraries. See https://github.com/huggingface/moon-landing/pull/4071.

PR #426, opened 2022-10-18
Ok let's do only one source of truth for licenses

julien-c:
(ie i'll use this list from moon)


osanseviero:
With the region tag can we programmatically get the licenses section in moon? :exploding_head:

julien-c:
@osanseviero yes! moon PR to come

osanseviero:
Very cool!

Issue #275, opened 2022-08-04
Tracking integration for Video Classification


osanseviero:
## Tracking integration of task - Video Classification

Example discussion - https://huggingface.co/keras-io/video-transformers/discussions/1#62e9694f1ee086f609b4a500

Note that you're not expected to do all of the following steps. This PR helps track all the steps required to get a new task fully supported in the Hub 🔥

- [ ] Integration with Inference API. Select at least one of the following:
  - [ ] Added a `transformers` `pipeline`
  - [ ] Added to Community Inference API for 3rd party library
  - [ ] Added to Community Inference API for generic
- [ ] Added basic UI elements (icon, order specification, etc)
- [ ] Added a widget

Integration guide: https://hf.co/docs/hub/models-tasks

NielsRogge:
Would be great to add basic UI elements for it, as we now have video models as seen here: https://huggingface.co/models?other=video-classification

osanseviero:
cc @NimaBoscarino or @nateraw would any of you be interested in working on this? This corresponds to the UI elements section at https://huggingface.co/docs/hub/models-tasks#ui-elements

osanseviero:
Friendly ping

NimaBoscarino:
I wish, but I don't have the bandwidth for it right now 😞

osanseviero:
Now that we have the UI elements and the pipeline, we could add this to the inference API for transformers + add a widget. WDYT @Narsil @OlivierDehaene @mishig25 ?

Narsil:
https://github.com/huggingface/api-inference/pull/1252

osanseviero:
Very cool! The only missing step would be the widget

osanseviero:
@nateraw would you be up for updating https://huggingface.co/tasks/video-classification to use the new `pipeline`?

PR #431, opened 2022-10-18
Add video classification task

nateraw:
Adds video classification as supported task.

Related convo: https://github.com/huggingface/hub-docs/issues/275

Haven't viewed this in moon landing yet, so leaving as draft until I do.


nateraw:
Here it is in browser

Screen Shot 2022-10-18 at 6 28 25 PM

osanseviero:
I think this is just using visual input, so I would leave in CV for now (although I could also see a case for multimodal if we use the audio as an input as well). No strong opinion

julien-c:
icon looks good to me but still tagging @gary149

osanseviero:
Should we go ahead with this PR or @gary149 you wanna take a look?

gary149 (approved):
LGTM

PR #433, opened 2022-10-19
💙 Swapped tapex with tapas

merveenoyan:
As discussed in slack, changed the model of table QA from tapex to fine-tuned tapas.


osanseviero (approved):
Thanks!

PR #438, opened 2022-10-20
👷 Add action to lint code

coyotte508:
This is a manual action. Could be done automatically but it would probably annoy people.

Unfortunately not possible to test unless it's merged in `main`.

### DIFF
--- a/.github/workflows/manual_lint.yml
+++ b/.github/workflows/manual_lint.yml
@@ -1,0 +1,33 @@
+name: Manual lint
+
+# Could do on:pull_request with path on js/** to do it automatically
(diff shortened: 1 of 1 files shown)


coyotte508:
They would need to go on https://github.com/huggingface/hub-docs/actions, select the action & run it, and only maintainers with write access can run it I think.

julien-c:
Yep, i think that's a fine solution. @merveenoyan WDYT? (i think it was you who prompted this?)

merveenoyan:
@julien-c sorry I missed this mention. I think it's perfectly fine!

PR #439, opened 2022-10-20
add code to languages list of datasets

loubnabnl:
Add `code` to the list of languages displayed on the Hub for datasets.

### DIFF
--- a/js/src/lib/interfaces/Language.ts
+++ b/js/src/lib/interfaces/Language.ts
@@ -155,6 +155,11 @@ export const LANGUAGES: Record<string, Language> = {
 		name:       "Corsican",
 		nativeName: "corsu",
 	},
+	code: {
+		code:       "code",
+		name:       "Code",
+		nativeName: "code",
+	},
 	cr: {
 		code:       "cr",
 		name:       "Cree",


osanseviero (approved):
Looks good, thanks!

Although I do wonder if here we want to have a different tag for each programming language or a single one is fine. WDT?

cc @julien-c @lvwerra

julien-c (changes requested):
I don't think adding `code` here will solve the original issue actually (can you link to the original issue? That can be an internal Slack link, that's fine)

julien-c:
this list is for human languages only

osanseviero:
https://huggingface.slack.com/archives/C02EMARJ65P/p1666255575685159

osanseviero:
We have 48 datasets with code tag https://huggingface.co/datasets?language=language:code but code is not showed in the language list.

loubnabnl:
It could be nice to have separate programming languages but not sure if it makes sense to mix them with natural languages, also the existing code datasets only have one code tag so maybe it's fine to just add `code`

julien-c:
Closing this PR in favor of a better fix in internal repo!

PR #441, opened 2022-10-20
`ModelPipelineIcon` => `PipelineIcon`

julien-c:
To reflect the fact that they're used also for datasets (and spaces one day?)


julien-c:
I almost renamed to `TaskTag` and `TaskIcon` but i thought this was not super necessary and maybe cognitively simpler to keep a similar name. If you feel strongly about it happy to reconsider

osanseviero (approved):
LGTM

PR #455, opened 2022-10-27
Micro-optimization to make @coyotte508 happy 😂

julien-c:


osanseviero (approved):

julien-c:
what if other consumers of this library want to use the Set directly? =)

seems innocuous to me, merging

PR #456, opened 2022-10-28
Show intel cpu load info

mishig25:


Michellehbn (approved):
LGTM! Thanks so much!

VioletteLepercq:
Thanks!!

VioletteLepercq (commented):
Thanks! LGTM

julien-c:
i would also still mention CPU no?

js/src/lib/components/InferenceWidget/shared/WidgetInfo/WidgetInfo.svelte
```diff
@@ -19,6 +19,19 @@
...
+	function getStatusReport(
+		modelLoadInfo: ModelLoadInfo,
+		statuses: Record<LoadingStatus, string>
+	): string {
+		if (
+			modelLoadInfo.compute_type === "cpu" &&
+			modelLoadInfo.status === "loaded"
+		) {
+			return `The model is loaded and running on <a class="underline" href="https://huggingface.co/intel" target="_blank">Intel Xeon Ice Lake</a>.`;
+		}
```
julien-c, line 32: we could also display when it's running on GPU/TPU something else no? (in another PR)

mishig25:
> i would also still mention CPU no?

@Michellehbn could you help with the text 😃

js/src/lib/components/InferenceWidget/shared/WidgetInfo/WidgetInfo.svelte
mishig25: if I am not mistaken, currently api-interence https://github.com/huggingface/api-inference/pull/1183 outputs `compute_type` as either `cpu` or `undefined`.

I can make it show `gpu` when `compute_type` is `undefined`. Or should I ask api-interence team to explicitly provide what type of `compute_type` is ?

cc: @Narsil @OlivierDehaene

julien-c: would sound good to me

PR #474, opened 2022-11-04
Fix textgen widget warning

mishig25:


js/src/lib/components/InferenceWidget/widgets/TextGenerationWidget/TextGenerationWidget.svelte
```diff
@@ -135,8 +135,12 @@
 				warning = "No text was generated";
 			} else if (model?.pipeline_tag === "text-generation") {
 				const outputWithoutInput = output.slice(text.length);
-				inferenceTimer.stop();
-				await renderTypingEffect(outputWithoutInput);
+				if(outputWithoutInput.length === 0){
+					warning = "No text was generated";
+				}else{
+					inferenceTimer.stop();
```
thomasw21, line 141: Do you not need to stop the timer regardless of the generated text?

PR #479, opened 2022-11-06
Fix licenses table

julien-c:


osanseviero (approved):
Thanks! Are the changes in zero-shot-classification due to linter?

PR #507, opened 2022-11-16
Revert "Update pr doc build action (#483)"

mishig25:
This reverts commit e2f44a95116add7308c1f862d347e64bc31b517d.


osanseviero (approved):
Approving if that unblocks you, but why is this getting reverted?

mishig25:
because it was using a "test branch" of doc-builder for testing the migration. 
Now that we checked everything works, it is back to using "main" branch

PR #512, opened 2022-11-17
[Paddle Integration] Add Widgets for PaddleNLP

sijunhe:


js/src/lib/interfaces/Libraries.ts
```diff
@@ -148,6 +149,28 @@ const keras = (model: ModelData) =>
...
+		].join("\n");
+	} else {
+		return [
+			`from paddlenlp.transformers import ${info.auto_model}`,
+			"",
+			`model = ${info.auto_model}.from_pretrained("${model.id}"${model.private ? ", use_auth_token=True" : ""}, from_hf_hub=True)`,
+		].join("\n");
+	}
+};
+
```
osanseviero, line 173: To simplify this PR as the information from the PaddleNLP is not in the `transformersInfo` as the config/library is different, would it be possible to do a very high-level snippet and afterward we could increase the coverage? First, we need this PR merged to start working on the internal side of things to add more info to the `ModelData`. Something along these lines


```suggestion
const paddlenlp = (model: ModelData) => {
	return [
	  `from paddlenlp.transformers import AutoModel`,
	  "",
	  `model = AutoModel.from_pretrained("${model.id}"${model.private ? ", use_auth_token=True" : ""}, from_hf_hub=True)`,
        ].join("\n");
};

```

osanseviero (commented):
Thanks a lot for the PR

js/src/lib/interfaces/Libraries.ts
sijunhe: Thanks for the suggestion. I basically took your suggestion. Only added a `AutoTokenizer`

osanseviero (approved):
Thanks a lot! 🔥

sijunhe:
Thanks for the review @osanseviero. Are we merging this? I don't have write access so I can't merge

osanseviero:
Yes, merged!

PR #518, opened 2022-11-18
chore: reduce the code example.

sayakpaul:
The inference code shown in https://huggingface.co/tasks/video-classification takes most of the place on the task page. It might be intimidating for beginners, which is why I reduced the inference code. 

Thanks to @osanseviero for reporting this.


merveenoyan (approved):
LGTM

PR #532, opened 2022-11-30
document upgraded spaces billing

christophe-rannou:


docs/hub/spaces-gpus.md
```diff
@@ -77,3 +77,19 @@ import tensorflow as tf
 print(tf.config.list_physical_devices('GPU'))
 # [PhysicalDevice(name='/physical_device:GPU:0', device_type='GPU')]
 ```
+
+## Billing
+
+Billing on Spaces is based on hardware usage and is minute based.
+
+During its lifecycle a Space is only billed when the Space is actually `Running` meaning that there is no cost during either build or startup.
```
osanseviero, line 85:
```suggestion
During a Space's lifecycle, it is only billed when the Space is actually `Running`. This means that there is no cost during either build or startup.
```

docs/hub/spaces-gpus.md
```diff
@@ -77,3 +77,19 @@ import tensorflow as tf
...
+
+If during runtime a space starts repeatedly crashing it is automatically suspended hence suspending the billing as well since the Space is no longer `Running`.
```
osanseviero, line 87:
```suggestion
If a running Space starts to fail, it will be automatically suspended, hence the billing will stop.
```

docs/hub/spaces-gpus.md
```diff
@@ -77,3 +77,19 @@ import tensorflow as tf
...
+
+Contrary to Spaces using free hardware a Space with upgraded hardware is not suspended automatically if it is not used during an extended period of time. If you wish to interrupt billing on your Space you will need to change its hardware settings back to CPU basic.
```
osanseviero, line 89:
```suggestion
Spaces running on free hardware are suspended automatically if they are not used for an extended period of time (e.g. two days). Upgraded Spaces will run indefinitely, even if there is no usage. To interrupt the billing on your Space, you can change the Hardware to CPU basic.
```

docs/hub/spaces-gpus.md
```diff
@@ -77,3 +77,19 @@ import tensorflow as tf
...
+
+<Tip>
+Additional lifecycle control over Spaces with upgraded hardware will soon be provided such as configuration of an automated suspension of the Space after a custom delay.
+</Tip>
```
osanseviero, line 93: I think it's ok to remove this and update the doc in few days/weeks

docs/hub/spaces-gpus.md
```diff
@@ -77,3 +77,19 @@ import tensorflow as tf
...
+
+Additional information about can be found in the [dedicated section](./billing).
```
osanseviero, line 95:
```suggestion
Additional information about billing can be found in the [dedicated section](./billing).
```

osanseviero (approved):
Very nice!

docs/hub/spaces-gpus.md
apolinario: I kind of like teasing this

PR #540, opened 2022-12-05
Put conversational live

merveenoyan:
Make this go brrr 
Ekran Resmi 2022-12-05 14 27 14

### DIFF
--- a/tasks/src/conversational/about.md
+++ b/tasks/src/conversational/about.md
@@ -8,10 +8,6 @@ Chatbots are used to have conversations instead of providing direct contact with
 
 Conversational response models are used as part of voice assistants to provide appropriate responses to voice based queries. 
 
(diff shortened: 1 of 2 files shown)


julien-c:
Looks good!

For models a seminal one on HF is `msft/dialogpt`
For datasets maybe `dailydialog`

osanseviero:
@julien-c there are already models and datasets defined in https://github.com/huggingface/hub-docs/blob/main/tasks/src/conversational/data.ts, Merve does not have a copy of the database, so it showed as if there were no models.

merveenoyan:
@osanseviero addressed your comment 🤗

merveenoyan:
@julien-c FYI the DB doesn't exist on codespaces so whenever I post it it will look empty 😅

PR #549, opened 2022-12-09
Add LIBRARY_TASK_MAPPING_EXCLUDING_TRANSFORMERS

xianbaoqian:
Context:  https://huggingface.slack.com/archives/C02ERUY9TKK/p1669743642830359

### DIFF
--- a/js/src/lib/interfaces/Types.ts
+++ b/js/src/lib/interfaces/Types.ts
@@ -749,3 +749,97 @@ export interface TransformersInfo {
 	 */
 	processor?: string;
 }
+
+
+/**
+ * Mapping from library name (excluding Transformers) to its supported tasks. 
+ * Inference API should be disabled for all other library, task pairs beyond this mapping.
+ * As an exception, Transformers supports all inference tasks.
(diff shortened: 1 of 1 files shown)


js/src/lib/interfaces/Types.ts
```diff
@@ -749,3 +749,97 @@ export interface TransformersInfo {
 	 */
 	processor?: string;
 }
+
+
+/**
+ * Mapping from library name (excluding Transformers) to its supported tasks. 
+ * Inference API should be disabled for all other library, task pairs beyond this mapping.
```
osanseviero, line 756:
```suggestion
 * Inference API should be disabled for all other (library, task) pairs beyond this mapping.
```

js/src/lib/interfaces/Types.ts
```diff
@@ -749,3 +749,97 @@ export interface TransformersInfo {
...
+ * As an exception, Transformers supports all inference tasks.
```
osanseviero, line 757:
```suggestion
 * As an exception, we assume Transformers supports all inference tasks.
```

osanseviero (approved):
Very nice!

xianbaoqian:
Thanks Omar for the detailed review!

PR #552, opened 2022-12-12
✨ Better typing for LIBRARY_TASK_MAPPING_EXCLUDING_TRANSFORMERS

SBrandeis:
Following #549 
Better / stricter typing, allowing only valid ModelLibrary tags and valid PipelineType

### DIFF
--- a/js/src/lib/interfaces/Libraries.ts
+++ b/js/src/lib/interfaces/Libraries.ts
@@ -32,7 +32,8 @@ export enum ModelLibrary {
 	"pythae"                 = "Pythae",
 }
 
-export const ALL_MODEL_LIBRARY_KEYS = Object.keys(ModelLibrary) as (keyof typeof ModelLibrary)[];
+export type ModelLibraryKey = ModelLibraryKey;
+export const ALL_MODEL_LIBRARY_KEYS = Object.keys(ModelLibrary) as ModelLibraryKey[];
 
 
 /**
@@ -390,7 +391,7 @@ model = AutoModel.load_from_hf_hub("${model.id}")`;
 
 
 
-export const MODEL_LIBRARIES_UI_ELEMENTS: { [key in keyof typeof ModelLibrary]?: LibraryUiElement } = {
+export const MODEL_LIBRARIES_UI_ELEMENTS: { [key in ModelLibraryKey]?: LibraryUiElement } = {
 	// ^^ TODO(remove the optional ? marker when Stanza snippet is available)
 	"adapter-transformers": {
 		btnLabel: "Adapter Transformers",
--- a/js/src/lib/interfaces/Libraries.ts
+++ b/js/src/lib/interfaces/Libraries.ts
@@ -32,7 +32,7 @@ export enum ModelLibrary {
 	"pythae"                 = "Pythae",
 }
 
-export type ModelLibraryKey = ModelLibraryKey;
+export type ModelLibraryKey = keyof typeof ModelLibrary;
 export const ALL_MODEL_LIBRARY_KEYS = Object.keys(ModelLibrary) as ModelLibraryKey[];
 
 
--- a/js/src/lib/interfaces/Libraries.ts
+++ b/js/src/lib/interfaces/Libraries.ts
@@ -9,10 +9,12 @@ export enum ModelLibrary {
 	"allennlp"               = "allenNLP",
 	"asteroid"               = "Asteroid",
 	"diffusers"              = "Diffusers",
+	"doctr"                  = "docTR",
 	"espnet"                 = "ESPnet",
 	"fairseq"                = "Fairseq",
 	"flair"                  = "Flair",
 	"keras"                  = "Keras",
+	"k2"                     = "K2",
 	"nemo"                   = "NeMo",
 	"paddlenlp"              = "PaddleNLP",
 	"pyannote-audio"         = "pyannote.audio",
@@ -34,6 +36,10 @@ export enum ModelLibrary {
 
 export type ModelLibraryKey = keyof typeof ModelLibrary;
 export const ALL_MODEL_LIBRARY_KEYS = Object.keys(ModelLibrary) as ModelLibraryKey[];
+
+const EXCLUDE_THOSE_LIBRARIES_FROM_DISPLAY: ModelLibraryKey[] = ["doctr", "k2"];
+
+export const ALL_DISPLAY_MODEL_LIBRARY_KEYS = ALL_MODEL_LIBRARY_KEYS.filter(k => !EXCLUDE_THOSE_LIBRARIES_FROM_DISPLAY.includes(k));
 
 
 /**
--- a/js/src/lib/interfaces/Libraries.ts
+++ b/js/src/lib/interfaces/Libraries.ts
@@ -397,7 +397,7 @@ model = AutoModel.load_from_hf_hub("${model.id}")`;
 
 
 
-export const MODEL_LIBRARIES_UI_ELEMENTS: { [key in ModelLibraryKey]?: LibraryUiElement } = {
+export const MODEL_LIBRARIES_UI_ELEMENTS: Partial<Record<ModelLibraryKey, LibraryUiElement>> = {
 	// ^^ TODO(remove the optional ? marker when Stanza snippet is available)
 	"adapter-transformers": {
 		btnLabel: "Adapter Transformers",
--- a/js/src/lib/interfaces/Types.ts
+++ b/js/src/lib/interfaces/Types.ts
@@ -1,4 +1,4 @@
-import type { ModelLibrary } from "./Libraries";
+import type { ModelLibraryKey } from "./Libraries";
 
 // Warning: order of modalities here determine how they are listed on the /tasks page
 export const MODALITIES = [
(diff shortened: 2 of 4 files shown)


js/src/lib/interfaces/Types.ts
```diff
@@ -758,88 +760,88 @@ export interface TransformersInfo {
...
 	"diffusers": [
-		"text-to-image"
-	],
-	"doctr": [
-		"object-detection"
+		"text-to-image",
 	],
+	// "doctr": [
+	// 	"object-detection",
+	// ],
```
SBrandeis, line 781: Commented because it's absent from `ModelLibrary`

SBrandeis: Same for superb & k2_sherpa

julien-c (commented):
lgtm

js/src/lib/interfaces/Types.ts
```diff
@@ -1,3 +1,5 @@
+import type { ModelLibrary } from "./Libraries";
```
julien-c, line 1: i would maybe put `LIBRARY_TASK_MAPPING_EXCLUDING_TRANSFORMERS` into a new file to prevent the circular imports between `Libraries.ts` and `Types.ts` (even if it's just types)

julien-c: like for the already existing `TASKS_MODEL_LIBRARIES` which is very similar, BTW (cc @xianbaoqian)

js/src/lib/interfaces/Types.ts
coyotte508: Should it be handled from the source huggingface/api-inference-community? (see  https://github.com/huggingface/api-inference-community/pull/158)

xianbaoqian: @osanseviero Any thoughts how to handle the discrepancy. Is the inference API working for doctr, superb & k2_sherpa? Or do we need to add these to `ModelLibrary`?

I didn't find anything related to k2_sherpa in the hub. I had a quick glimpse on a few superb models and they seem to be using transformer for inference API (?) Most doctr repos are dummy models and the inference API is unable to decide the pipeline_tag.

xianbaoqian:
Thanks for making it strong typed! Left a comment.

js/src/lib/interfaces/Types.ts
osanseviero: > Is the inference API working for doctr, superb & k2_sherpa? Or do we need to add these to ModelLibrary?

* Superb repos use the Community Inference API, but given they are not a library and their usage is more for a competition, they were not added to the `Libraries.ts` (confirmed with @lewtun). From a discussion, we think https://huggingface.co/models?library=superb&sort=modified are mostly old legacy models and it's ok to remove them from the Inference API, as the official ones now use `transformers`. (the old integration allowed users to specify their own requirements and inference code but that was deprecated)
*  I don't have context in the `k2-sherpa` integration, so maybe makes sense to remove entirely from the community API or exclude from the exporter
* https://huggingface.co/models?library=doctr&sort=modified is just one model so I'm fine removing `doctr`

lewtun: Agree that it's OK to remove all the models listed here: https://huggingface.co/models?library=superb&sort=modified

(These were just for testing)

osanseviero: @xianbaoqian would you be up for opening a PR to exclude these? :fire:

xianbaoqian: Sure. Will update [api-inference-community](https://github.com/huggingface/api-inference-community) to remove these.

xianbaoqian: I'm a bit surprised that using library: k2 actually made the inference API working. I thought that it needs the same name as the folder name in api-inference-community. 

Can anyone share a pointer on how they communicate? Not sure if it's easier to just update the repo library name from k2 to k2_sherpa, or to rename k2_sherpa folder to k2. Either way, we also need to update the library constant in the .ts file.

osanseviero: The `ModelInfo` `library_name` is not enforced/validated, so it can be anything. When sending requests to the Inference API, it will pick the library from this internal list (https://github.com/huggingface/api-inference/blob/main/master/app/types.py#L21), which, as you can see, contains some libraries that are not in `Libraries.ts` such as `doctr`, `superb`, `generic`, etc. 

**We do want a mechanism in which we want to run Inference API on new libraries without adding a filter at the left of hf.co/models though. Making this strongly typed would disallow us from doing so. cc @SBrandeis any ideas on this? **

xianbaoqian: > We do want a mechanism in which we want to run Inference API on new libraries without adding a filter at the left of hf.co/models though.

Thanks for the information @osanseviero  I think both k2 and doctr could be good examples. They are only limited number of models available so it makes sense not showing them on the UI in the models page but it's probably good to support them on the inference API given the efforts that we put in to support them in the first place. (also cc @Narsil who has been working on k2 integration)

I do like strong typed list over weak typed list for many reasons such as detecting typos much early on, etc. Two thoughts:
- Maybe we can just create a different constant list of keys for libraries that supports inference API.
- or we can follow the example in https://github.com/huggingface/hub-docs/blob/main/js/src/lib/interfaces/Types.ts#L609 and defines a `hideInModels` flag for libraries. If this flag is on, the library will not show up in filter on the page to the models page. (kudos to @osanseviero)

Either way, this issue is getting more complicated. Probably we want to move forward with https://github.com/huggingface/moon-landing/pull/4816 first before the strong typed thing is figured out?

Any thoughts @SBrandeis @julien-c ?

Narsil: All up for static typing ! 

I think the current issues are not worth stopping the current PR given the quality of life improvement static typing provides.

That being said, the fact that the mapping is not easily aligned currently just shows to me that `doctr` and `k2` are just missing from the library keys, no ? So instead of removing them couldn't we add them where they are missing ?

If the issue would be making them visible as filters, then maybe we're just lacking an exclusion list from the filter in the search.

julien-c: IMO the simplest way is add `k2` and `doctr` (and any other future library implemented in api-inference) to the `ModelLibrary` type, but add a blocklist (or allow-list) of libraries that are not displayed in /models

Let me push a proposal

julien-c: 64e8088c19cc80de897a3eb7393507db54e77f60

js/src/lib/interfaces/Types.ts
```diff
@@ -758,88 +760,84 @@ export interface TransformersInfo {
...
+		"token-classification",
 	],
 	"allennlp": [
-		"question-answering"
+		"question-answering",
 	],
 	"asteroid": [
-		"audio-source-separation",
-		"audio-to-audio"
+		// "audio-source-separation",
```
julien-c, line 773: (not a valid pipeline tag)

js/src/lib/interfaces/Types.ts
julien-c: for me this PR is ready to merge

julien-c (approved):

js/src/lib/interfaces/Types.ts
xianbaoqian: Thanks! I think adding the two definitely makes sense. 

Another issue is that people have been using k2 in HuggingFace tags and k2_sherpa in api-inference-community. I think the easiest fix is to rename k2_sherpa to k2. I'll propose a PR in api-inference-community!

js/src/lib/interfaces/Types.ts
julien-c: (we can change that in the future though – no need to do it in this PR)

js/src/lib/interfaces/Types.ts
julien-c: (also see internal PR https://github.com/huggingface/moon-landing/commit/e9557f8c848abbb98e48e8c4023ac6526022273e)

julien-c: and yes renaming k2_sherpa to k2 sounds perfect

osanseviero: 64e8088 was exactly what I had in mind when we discussed @xianbaoqian. :) Thanks for doing it @julien-c!

js/src/lib/interfaces/Libraries.ts
```diff
@@ -390,7 +397,7 @@ model = AutoModel.load_from_hf_hub("${model.id}")`;
 
 
 
-export const MODEL_LIBRARIES_UI_ELEMENTS: { [key in keyof typeof ModelLibrary]?: LibraryUiElement } = {
+export const MODEL_LIBRARIES_UI_ELEMENTS: { [key in ModelLibraryKey]?: LibraryUiElement } = {
```
osanseviero, line 400: Will this cause issues for `doctr` and `k2` which will be displayed in the model page but the key does not exist?

osanseviero (approved):
Nice, overall looks good!

SBrandeis:
Thanks for handling this @julien-c 🤗

js/src/lib/interfaces/Libraries.ts
julien-c: i don't think so because for instance Stanza (and I'm sure others) currently don't have a `LibraryUiElement`

julien-c: oh actually, scratch that, it does

julien-c: but i think it should be fine, and btw we can maybe update the type the clearer:

```suggestion
export const MODEL_LIBRARIES_UI_ELEMENTS: Partial<Record<ModelLibraryKey, LibraryUiElement>> = {
```

SBrandeis: The only place I see this variable used is [internal link](https://github.com/huggingface/moon-landing/blob/main/server/views/components/ModelLibraryModal/ModelLibraryModal.svelte)

And it seems to be already handling missing library names gracefully, so I guess it's fine?

osanseviero: Perfect then!

PR #567, opened 2022-12-16
Adds support for OpenCLIP

LysandreJik:
Adds support for OpenCLIP as a library with a code snippet.

cc @rwightman

### DIFF
--- a/js/src/lib/interfaces/Libraries.ts
+++ b/js/src/lib/interfaces/Libraries.ts
@@ -16,7 +16,7 @@ export enum ModelLibrary {
 	"keras"                  = "Keras",
 	"k2"                     = "K2",
 	"nemo"                   = "NeMo",
-	"openclip"               = "OpenCLIP",
+	"open_clip"               = "OpenCLIP",
 	"paddlenlp"              = "PaddleNLP",
 	"pyannote-audio"         = "pyannote.audio",
 	"sample-factory"         = "Sample Factory",
@@ -158,7 +158,7 @@ const keras = (model: ModelData) =>
 model = from_pretrained_keras("${model.id}")
 `;
 
-const openclip = (model: ModelData) =>
+const open_clip = (model: ModelData) =>
 	`import open_clip
 
 model, preprocess_train, preprocess_val = open_clip.create_model_and_transforms('hf-hub:${model.id}')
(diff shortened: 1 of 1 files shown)


osanseviero:
(this is cool!)

For usage of `hf_hub_download` in openclip, setting `library_name` and `library_version` is also nice :)

LysandreJik:
Thanks for the feedback, I'll add that to the OpenCLIP PR!

rwightman:
Looks good, 

@osanseviero you mean? https://github.com/mlfoundations/open_clip/blob/main/src/open_clip/pretrained.py#L14

osanseviero:
Yes, maybe https://github.com/mlfoundations/open_clip/blob/main/src/open_clip/pretrained.py#L282 should also change but I'm not very familiar with the codebase

rwightman:
> Yes, maybe https://github.com/mlfoundations/open_clip/blob/main/src/open_clip/pretrained.py#L282 should also change (which is the function that made me say the comment above). but I'm not very familiar with the codebase

It's using the fn I linked to above, it bound the lib/version args

julien-c:
i don't think `library_name` and `library_version` are used in Elasticsearch stats currently (unless i'm mistaken?)

osanseviero:
Yes, this does not impact download stats but this should allow us to do fine grained analysis at any point in the future if there's any interest in that. We have this in all the integrations.

julien-c:
yes i was only commenting on @rwightman's question 
> if there are a decent number of existing downloads [based on this]

rwightman:
we should probably line up the names across lib name and tags either way? but good to know re elasticsearch stats.

LysandreJik:
Updated the code snippet to also feature the tokenizer and updated `openclip` to `open_clip` to be coherent.

Switching this to non-draft mode as https://github.com/mlfoundations/open_clip/pull/375 was merged.

js/src/lib/interfaces/Libraries.ts
```diff
@@ -16,6 +16,7 @@ export enum ModelLibrary {
 	"keras"                  = "Keras",
 	"k2"                     = "K2",
 	"nemo"                   = "NeMo",
+	"open_clip"               = "OpenCLIP",
```
Pierrci, line 19:
```suggestion
	"open_clip"              = "OpenCLIP",
```
😃

Pierrci (commented):
LGTM

js/src/lib/interfaces/Libraries.ts
LysandreJik: oops, good catch! Thanks @Pierrci

coyotte508 (approved):

LysandreJik:
Thanks for the reviews @Pierrci and @coyotte508!

PR #566, opened 2022-12-16
Correct typo in spaces-github-actions.md

javierlorenzod:
HF_USERNAME is not included in the space's URL

PR #579, opened 2022-12-21
Update spaces-github-actions.md

javierlorenzod:
- Separate FULL_SPACE_NAME into HF_USERNAME and SPACE_NAME, as suggested in #566 by @julien-c


osanseviero (approved):
SGTM, WDYT @julien-c ?

julien-c (approved):
yep i think that's better

PR #580, opened 2022-12-22
refactor: Move LIBRARY_TASK_MAPPING_EXCLUDING_TRANSFORMERS closer to TASKS_MODEL_LIBRARIES

xianbaoqian:
From the review comment by Julien. Context: https://github.com/huggingface/hub-docs/pull/552#discussion_r1045843182

Although semantically the two maps should be the same and one should generate another, but in practice, there are a few differences. 

IIUC, TASKS_MODEL_LIBRARIES is only used for task page and it seems to be manually maintained.
LIBRARY_TASK_MAPPING_EXCLUDING_TRANSFORMERS is used for inference and is auto-generated from api-inference-community.

This PR is only about refactoring the existing code, not about merging the two together. Just a few thoughts on the merging:

I think the latter is more reliable and probably more up-to-date also, but not sure if we will have a case where we can show a library on the task page, even if the inference API is not ready. Also the latter doesn't have any information on transformers and assume that transformers can handle all tasks, which is not really the case per the former, so merging the two is not enough unless we have a new list of tasks that transformers can't handle :-)


Pierrci:
I think the `import type { ModelLibraryKey } from "./Libraries";` in `Types.ts` can then be removed?

xianbaoqian:
Thanks @Pierrci  ! Removed.

tasks/src/const.ts
```diff
@@ -46,3 +46,93 @@ export const TASKS_MODEL_LIBRARIES: Record<PipelineType, ModelLibraryKey[]> = {
...
 };
+
+/**
+ * Mapping from library name (excluding Transformers) to its supported tasks. 
+ * Inference API should be disabled for all other (library, task) pairs beyond this mapping.
+ * As an exception, we assume Transformers supports all inference tasks.
+ * This mapping is generated automatically by "python-api-export-tasks" action in huggingface/api-inference-community repo upon merge.
+ * Ref: https://github.com/huggingface/api-inference-community/pull/158
+ */
+export const LIBRARY_TASK_MAPPING_EXCLUDING_TRANSFORMERS: Partial<Record<ModelLibraryKey, PipelineType[]>> = {
```
julien-c, line 57: can you move it to its own file, i.e. a separate new file, like `./src/lib/interfaces/LibrariesToTasks.ts`?

the `/tasks` folder is more specific to the `/tasks` pages on the Hub, as you mention

julien-c (commented):
Oh yeah i was not suggesting that you merge the two mappings, just that they were similar.

i'd rather have the mapping in a file by itself (see my suggestion)

Thanks!

tasks/src/const.ts
xianbaoqian: Cool. Done.

julien-c (approved):
looks good!

PR #584, opened 2022-12-27
Add an Example using Julia's Genie.jl framework

anoojpatel:
I've created a demo using Julia lang's Genie.jl framework to build a webapp using Docker spaces! I would love to add this to the list of examples, as it is quite finicky to figure out how ports need to be exposed and how the default CMD is leveraged. I think this would be useful for those looking to grow the Julia ecosystem and support on the Huggingface platform :)


julien-c (approved):
Very cool! Would you be up for writing a short write-up about how the steps you went through, (as well as a quick intro to Genie.jl) which we could have directly in hf.co/docs/hub?

anoojpatel:
Yes definitely! Where should I put this specific write up?

julien-c:
maybe in `spaces-docker-julia-genie.md` file? (we can always rename later if needed)

julien-c:
in the meantime, i'm deploying this PR :) Thanks a lot!

julien-c:
see also #585 for a similar approach! (documenting a new SDK for Spaces)

apatel-fn:
Amazing! Thanks for the reference :)

PR #592, opened 2023-01-04
Ignore a11y warnings for newer svelte versions

krampstudio:
Preparing the update of Svelte `^3.55.1`

The newer version of Svelte comes with new a11y warnings. I just ignore them for now, but maybe we need to consider fixing some later (especially for keyboard support).


coyotte508 (approved):

mishig25 (approved):
lgtm !

PR #597, opened 2023-01-05
First version of "Tags that are suggested inside the metadata GUI"

julien-c:
is a first step to close [this internal issue](https://github.com/huggingface/moon-landing/issues/4381)

### DIFF
--- a/js/src/lib/interfaces/Types.ts
+++ b/js/src/lib/interfaces/Types.ts
@@ -664,6 +664,25 @@ export const PIPELINE_TAGS_DISPLAY_ORDER: Array<PipelineType> = [
 	"depth-estimation",
 ];
 
+/**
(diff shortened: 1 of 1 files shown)


js/src/lib/interfaces/Types.ts
```diff
@@ -664,6 +664,25 @@ export const PIPELINE_TAGS_DISPLAY_ORDER: Array<PipelineType> = [
...
+ * Let's keep their number not too high to not spread out models/datasets too much.
+ */
+export const OTHER_TAGS_SUGGESTIONS = [
+	"biomedical",
+	"chemistry",
+	"finance",
+	"legal",
+	"music",
+	"science",
+	"waifu-diffusion",
```
julien-c, line 682: was not sure about this last one...

PR #598, opened 2023-01-05
Some changes of the default shown tags

osanseviero:


julien-c (approved):
i would add a `medical` or `health`, other than that, LGTM

PR #604, opened 2023-01-11
added instance segmentation task spaces

rashmibanthia:


tasks/src/image-segmentation/data.ts
```diff
@@ -59,6 +59,10 @@ const taskData: TaskDataCustom = {
 			description: "A semantic segmentation application that predicts human silhouettes.",
 			id:          "keras-io/Human-Part-Segmentation",
 		},
+		{
+			description: "Cell Instance Segmentation - Sartorius Kaggle Competition",
```
merveenoyan, line 63: Can you give more simpler context on what this is about? You can take a look at the other definitions 🙂

merveenoyan (changes requested):
Thanks a lot for this PR, I just left a comment, please take a look 🙂

rashmibanthia:
Does this description sound any better or more complex :D ? 
“An instance segmentation application to predict neuronal cell types from microscopy images. (Sartorius Kaggle Competition)”

merveenoyan:
@rashmibanthia maybe we could remove kaggle part and it would be less cryptic IMO 😅

tasks/src/image-segmentation/data.ts
merveenoyan, line 63:
```suggestion
			description: "An instance segmentation application to predict neuronal cell types from microscopy images.",
```
Does this look good?

merveenoyan (commented):
@rashmibanthia left suggestion 🙂

tasks/src/image-segmentation/data.ts
rashmibanthia: Yes perfect, Thank you very much @merveenoyan.

PR #613, opened 2023-01-13
Removed tasks and added snippet

merveenoyan:
Removed voice activity recognition, robotics and text2text generation tasks from tasks front page and added a small code snippet to zero shot classification.


merveenoyan:
result: 
Ekran Resmi 2023-01-13 23 07 39
Ekran Resmi 2023-01-13 23 07 59

osanseviero (approved):
Nice!

julien-c:
i don't remember, why are we removing robotics and voice-activity-detection? (given they have filters on hf.co/models and hf.co/datasets)

osanseviero:
https://huggingface.co/models?pipeline_tag=robotics&sort=downloads now leads to 0 models. There were like 5 models before, but it seems they were removed. Also, from the perspective of a ML task, it's not a super clear one (it's more of a vertical I would say). I think we should hide it in the models search for now as well.

osanseviero:
And https://huggingface.co/models?pipeline_tag=voice-activity-detection&sort=downloads is just 6 models by pyannote. Do we think it's worth adding a task page to a task with such low volume?

merveenoyan:
@julien-c they're too niche and there's no models so once we have models we can add them back no?

julien-c:
okay, sounds good

julien-c (approved):

PR #614, opened 2023-01-14
Added yolov8/7/6/x models.

kadirnar:


tasks/src/object-detection/data.ts
```diff
@@ -60,6 +60,22 @@ const taskData: TaskDataCustom = {
 			description: "An application that contains various object detection models to try from.",
 			id:          "Gradio-Blocks/Object-Detection-With-DETR-and-YOLOS",
 		},
+		{
+			description: "Ultralytics YOLOv8: State-of-the-Art YOLO Models.",
```
merveenoyan, line 64: Is it okay if you could pick two of these Spaces instead, given most of them are similar? Maybe pick most distinct ones?

merveenoyan (commented):
It's very elaborate how you put the descriptions, thanks a lot! Left a small comment 🙂

tasks/src/object-detection/data.ts
kadirnar: The most recent YOLO repositories are YOLOv8 and YOLOv6.

kadirnar: I fixed it.

merveenoyan (approved):
LGTM!

PR #616, opened 2023-01-15
add capdec space in image-to-text

johko:
I added the space for CapDec ([Text-Only Training for Image Captioning using Noise-Injected CLIP](https://arxiv.org/pdf/2211.00575.pdf)) on the Image-to-Text task page

### DIFF
--- a/tasks/src/image-to-text/data.ts
+++ b/tasks/src/image-to-text/data.ts
@@ -63,7 +63,7 @@ const taskData: TaskDataCustom = {
 			id:          "Salesforce/BLIP",
 		},
 		{
-			description: "An image captioning application that demonstrates the effect of noise injections on captions.",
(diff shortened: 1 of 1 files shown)


tasks/src/image-to-text/data.ts
```diff
@@ -62,6 +62,10 @@ const taskData: TaskDataCustom = {
 			description: "An application that can caption images and answer questions about a given image.",
 			id:          "Salesforce/BLIP",
 		},
+		{
+			description: "An image captioning application that demonstrates the effect of noise injections on captions.",
```
merveenoyan, line 66:
```suggestion
			description: "An image captioning application that demonstrates the effect of noise on captions.",
```

merveenoyan (approved):
otherwise LGTM!

PR #622, opened 2023-01-19
Registering experimaestro IR library

bpiwowar:
Hi,

I am unsure how to proceed for this - and even how to test this PR, but this should PR goal is to add the [Experimaestro IR library](https://github.com/experimaestro/experimaestro-ir) designed for neural Information Retrieval experiments.


osanseviero (commented):
HI there! Thanks a lot for your PR! At the moment, there are only 3 models that have the `xpmir` tag. In order to add new libraries to the tags at hf.co/models, we usually wait for the library to have some dozens of models to add the library. We can revisit this PR later on if the usage increases a bit

julien-c:
closing as stale

PR #626, opened 2023-01-20
Fix snippet for zero-shot-classification for lang zh

mishig25:
Value for `candidate_labels` field needs to be string, not string[] as written [here](https://huggingface.co/docs/hub/models-widgets-examples#zeroshot-classification).
It should fix the issue for running widget on https://huggingface.co/PaddlePaddle/utc-large

cc: @sijunhe


sijunhe:
Nice. Thanks!

PR #634, opened 2023-01-25
added learning resources

ArafatAbdussalam:
I added to a link to Keras learning resources for developers who might want to take a refresher course before deep diving into Huggingface.


osanseviero (approved):
Thanks! :fire:

PR #638, opened 2023-01-25
Corrected 20 to 200, status code.

ankitjaininfo:
Also, added clarifications for local tunnels.


julien-c (changes requested):
Are you affiliated with Beeceptor? 🤣

I would not mention them again in the second paragraph as i think ngrok and localtunnel are more well-known for this use case

ankitjaininfo:
No I am not affiliated. I love their product and simplicity. When I saw you have added a section about localhost routing, I thought to mention it. I am push a change to remove Beeceptor.

Notice that, there is another correction for 20 ==> 200 status code.

julien-c (approved):
looks good to me! thanks

PR #639, opened 2023-01-26
Added tracking algorithm library.

kadirnar:
I added [torchyolo](https://huggingface.co/spaces/kadirnar/torchyolo) instead of [yolov6](https://huggingface.co/spaces/kadirnar/yolov6).

### DIFF
--- a/tasks/src/object-detection/data.ts
+++ b/tasks/src/object-detection/data.ts
@@ -65,7 +65,7 @@ const taskData: TaskDataCustom = {
 			id:          "kadirnar/yolov8",
 		},
 		{
-			description: "TorchYolo: YOLO Series Object Detection and Track Algorithm Library.",
+			description: "An application that shows multiple cutting edge techniques for object detection and tracking",
 			id:          "kadirnar/torchyolo",
 		},
 	],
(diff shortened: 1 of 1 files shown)


tasks/src/object-detection/data.ts
```diff
@@ -65,8 +65,8 @@ const taskData: TaskDataCustom = {
 			id:          "kadirnar/yolov8",
 		},
 		{
-			description: "YOLOv6: a single-stage object detection framework dedicated to industrial applications.",
-			id:          "kadirnar/yolov6",
+			description: "TorchYolo: YOLO Series Object Detection and Track Algorithm Library.",
```
merveenoyan, line 68: Can you write a more general explanation for this? 
Maybe something like:
"An application that shows multiple cutting edge techniques for object detection and tracking."

merveenoyan (commented):
Left a minor rephrase 🙂

tasks/src/object-detection/data.ts
kadirnar: Thank you. I fixed it.

merveenoyan (approved):

PR #641, opened 2023-01-28
Fix missing json module import

dkgv:
The `snippetFile` example is missing an import to the `json` module.

I have changed it to instead parse JSON using the response object like the other examples. It looks like decoding is taken care of by the requests library [automatically](https://github.com/psf/requests/blob/main/requests/models.py#L952-L968) anyway.


julien-c (commented):
lgtm, let me know when this is ready to merge and i'll merge!

dkgv:
I don't have anything further to add so you may merge whenever 👍

PR #645, opened 2023-01-30
Webook guide # 3 automatic metadata review

davanstrien:
A WIP guide on using webhooks for automatic metadata review. I've aimed to follow the style of the previous guides and only included some of the code used in the [example app](https://huggingface.co/spaces/librarian-bot/webhook_metadata_reviewer). 

I have added more context to the intro to motivate the example since this use case might be slightly less obvious than the previous guides. 

Still TODO:
- [ ] Add screenshots to https://huggingface.co/datasets/huggingface/documentation-images and images to docs

I thought it was best to get feedback on the content before doing the images, but once the images are done, I think this should be ready.

PR #647, opened 2023-01-31
Remove the YAML breaks from example

davanstrien:
Possibly YAML markers are breaking build of docs


julien-c:
(follow up to #645 )

For some reason now the doc build CI job fails 🤔

davanstrien:
Building no longer fails, but the preview still returns a 500 error for page added in #645

davanstrien:
> Building no longer fails, but the preview still returns a 500 error for page added in #645

I ran black on the code to enforce double quotes in the Python block to see if single quotes were causing issues, but this doesn't seem to have fixed the issue.

julien-c:
can be closed if #649 solves the issue

PR #649, opened 2023-02-01
Fix #645 breaking doc-builder

julien-c:
see this comment: https://github.com/huggingface/hub-docs/pull/645#issuecomment-1410816047

see double escaping of Bot inside quotes here:

Screenshot 2023-02-01 at 22 11 20

cc @mishig25 @coyotte508 for doc-builder


julien-c:
ok it's already tracked at https://github.com/huggingface/doc-builder/issues/343 (thanks @coyotte508)

PR #653, opened 2023-02-02
Put feature extraction live

merveenoyan:
Tabular schema was missing "label" so I added it to rest of the tabular tasks as well as feature extraction, and put feature extraction live.


osanseviero (approved):
Thanks!

PR #664, opened 2023-02-08
Enable download Git-LFS files on sync

radames:
Following an user issue, enable LFS on [sync](https://github.com/actions/checkout#:~:text=Whether%20to%20download%20Git%2DLFS%20files)

### DIFF
--- a/docs/hub/spaces-github-actions.md
+++ b/docs/hub/spaces-github-actions.md
@@ -35,6 +35,7 @@ jobs:
       - uses: actions/checkout@v3
         with:
           fetch-depth: 0
+          lfs: true
       - name: Push to hub
         env:
           HF_TOKEN: ${{ secrets.HF_TOKEN }}


NimaBoscarino:
What LFS files need to be downloaded for the build? AFAIK any image assets should be up on https://huggingface.co/datasets/huggingface/documentation-images, and I don't know if there are other binary files that would be part of the hub-docs. Holding off on approving, but otherwise this LGTM!

radames:
hi @NimaBoscarino , thanks for the comment. The changes is only for the GitHub Action example when a user wants to sync their GitHub repository with a Space, and the action need to explicitly enable sync LFS files in the repo.
An user, for example, had onnx models in their GitHub repo, and need to LFS push to our hub. It wasn't working without that option  `lfs: true` .

osanseviero (approved):
Thanks for this! cc @nateraw as well

NimaBoscarino:
@radames Oh LOL, I should have looked at the file name 🤦 I thought it was a GitHub action for this repo, hahahaha.

radames:
> @radames Oh LOL, I should have looked at the file name 🤦 I thought it was a GitHub action for this repo, hahahaha.
haha no worries, I was a bit confused but fun 😂

PR #665, opened 2023-02-09
Add reference to the base URL in the Hub API docs

apolinario:
It may feel obvious but we never state the actual URL to construct the API, which could confuse some users


osanseviero (approved):
Good catch! Thanks

PR #669, opened 2023-02-09
Add Space disk space info

osanseviero:
Related discussion https://huggingface.slack.com/archives/C048K60MPNF/p1675778705684199

### DIFF
--- a/docs/hub/spaces-gpus.md
+++ b/docs/hub/spaces-gpus.md
@@ -18,6 +18,20 @@ As soon as your Space is running on GPU you can see which hardware it’s runnin
 <img class="hidden dark:block" src="https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/hub/spaces-running-badge-dark.png"/>
 </div>
 
+## Hardware Specs
+
+In the following table, you can see the Specs for the different upgrade options.
+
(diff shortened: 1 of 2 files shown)


christophe-rannou (approved):
LGTM but is it me or there is an issue with the PR doc https://moon-ci-docs.huggingface.co/docs/hub/pr_669/en/spaces-overview ?
I can't see your changes there.

Edit: was not patient enough, I see the changes now !

julien-c (approved):
LGTM but we'll also need to make sure to keep those up to date :)

cc @SBrandeis who controls the ground truth for the prices

SBrandeis:
Would it make sens to display that information in the [pricing page](https://huggingface.co/pricing#spaces) instead, and maybe link to there from the docs?
It would avoid discrepancies between the docs and the actual prices / hardware specs, and keep moon-landing as the source of truth on that matter

osanseviero:
WDYT @julien-c? You mentioned nicer in the docs that in /settings, but what about /pricing?

julien-c:
no strong opinion – we can also merge them here for now, and think about it later

i.E. i think on /pricing for instance "disk space" might be a bit too much info, so i think it makes sense here

osanseviero:
Alright, merging ahead then!

PR #673, opened 2023-02-13
Add Widgets Support for MindSpore

hellowaywewe:


js/src/lib/interfaces/Libraries.ts
```diff
@@ -521,6 +534,12 @@ export const MODEL_LIBRARIES_UI_ELEMENTS: Partial<Record<ModelLibraryKey, Librar
 		repoUrl:  "https://github.com/fastai/fastai",
 		snippet:  fastai,
 	},
+	"mindspore": {
```
xianbaoqian, line 537: My understanding is that mindspore is the name of the framework, and tinyms is the name of the library.

Is all MindSpore model using tinyms, if so we can mix the two. Or do you have other libraries on top of MindSpore?

hellowaywewe: @xianbaoqian Thank you for your review. Yes, you are right, the mindspore is the name of the framework, and tinyms is the name of the library. All the MindSpore models can use tinyms, MindSopore's easy-to-use suite library, to adapt to HF to provide more AI experience services for users.
 I don't know what the repoUrl field in the Libraries.ts file is for, so I fill in tinyms to help the background download directly to tinyms to execute the task. If this field has no special technical purpose, It can be filled with the MindSpore framework name, I have updated, thank you.

xianbaoqian: I see. Thanks. I think it looks good now. 

Curious how do tinyms tells which the task of the model? Thx

xianbaoqian (approved):

js/src/lib/interfaces/Libraries.ts
hellowaywewe: TinyMS determines what tasks need to be done based on the network and the data set. For example, LeNet networks and mnist datasets perform the image-classification task. I don't know if I've answered your question.

js/src/lib/interfaces/Libraries.ts
```diff
@@ -294,6 +295,18 @@ const fastai = (model: ModelData) =>
 
 learn = from_pretrained_fastai("${model.id}")`;
 
+const mindspore = (model: ModelData) => {
+	const architecture = model.config.architectures[0].toLowerCase();
+	if (model.tags?.includes("image-classification")){
```
osanseviero, line 300: In case it's not an image classification model, we should have some fallback/unknown snippet (e.g. https://github.com/hellowaywewe/hub-docs/blob/main/js/src/lib/interfaces/Libraries.ts#L82). Just to confirm, the given snippet does not work unless it's a `image-classification` model?

osanseviero (commented):
Very cool PR! Sorry for the late review, I just have a small nit

js/src/lib/interfaces/Libraries.ts
hellowaywewe: @osanseviero Yes, you are right, currently we provide image-classification only as a test, but more tasks will be supported later. I have submitted a new PR, please review it. (https://github.com/huggingface/hub-docs/pull/673)

hellowaywewe: Sorry, this PR (https://github.com/huggingface/hub-docs/pull/677)

PR #674, opened 2023-02-13
spaces-docker: discourage use of chown or chmod

christophe-rannou:


julien-c (commented):
cool

also cc @cbensimon just in case

docs/hub/spaces-sdks-docker.md
```diff
@@ -72,10 +65,45 @@ ENV HOME=/home/user \
 # Set the working directory to the user's home directory
 WORKDIR $HOME/app
 
+# Try and run pip command after setting the user with `USER user` to avoid permission issues with Python
+RUN pip install --upgrade pip
```
XciD, line 69: --no-cache-dir

docs/hub/spaces-sdks-docker.md
```diff
@@ -72,10 +65,45 @@ ENV HOME=/home/user \
...
+# Download a checkpoint
+RUN mkdir content
+ADD --chown=user https://huggingface.co/<ORG>/<MODEL>/resolve/main/<CHECKPOINT>.ckpt content/<ORG>/<MODEL>/<CHECKPOINT>.ckpt
+```
+
+<Tip warning="{true}">
+Always specify the `--chown=user` with `ADD` and `COPY` to ensure the new files are owned by your user.
+</Tip>
+
+If you still face permission issues, you might need to use `chmod` or `chown` in your `Dockerfile` to grant the right permissions. For example, if you want to use the directory `/data`, you can do:
```
XciD, line 83: we should remove it IMO

christophe-rannou: I planned on that but that's required if the space is using folders like `/data` at the root, these cannot be created as a `user` so the need to be created then chown, and I feel it's better to show what to do in this instance

docs/hub/spaces-sdks-docker.md
```diff
@@ -72,10 +65,45 @@ ENV HOME=/home/user \
...
 
+# Try and run pip command after setting the user with `USER user` to avoid permission issues with Python
+RUN pip install --upgrade pip
+
 # Copy the current directory contents into the container at $HOME/app setting the owner to the user
 COPY --chown=user . $HOME/app
+
+# Download a checkpoint
+RUN mkdir content
+ADD --chown=user https://huggingface.co/<ORG>/<MODEL>/resolve/main/<CHECKPOINT>.ckpt content/<ORG>/<MODEL>/<CHECKPOINT>.ckpt
```
osanseviero, line 76: Not sure if we should show this, as in most cases we want people to use models through a downstream library and this might confuse users (e.g. `transformers`, `diffusers`, or even `huggingface_hub`)`

osanseviero (commented):
Looks nice!

PR #676, opened 2023-02-14
Add gradio docker example

johko:
Hey, I recently was looking for a docker space example that just uses Gradio and it was harder to find than I thought.
I thought it would be nice to include @sayakpaul's space here in the general examples.


osanseviero (commented):
Out of curiosity, why would a user prefer a Docker Space for Gradio rather than the native Gradio Space?

sayakpaul:
So, I hit on a likely upcoming feature accidentally? I will take the day off to celebrate 🎉

julien-c:
haha still interested in why you needed this, out of curiosity!

sayakpaul:
So, I was working on [this Space](https://t.co/yvb2OPQ9je) that lets you optimize a ResNet50 model. The Space required a Docker support (TensorRT can sometimes destroy your mood out of nowhere). I wanted the results to be presentable and that is why I ended up exploring this integration: Docker + Spaces + Gradio.

julien-c:
I see! that's a valid use case

johko:
In my case pytorch3d is the culprit. It is hell to set up (even locally) without docker. And I want to use it in a space.

To save some time and not despair completely I'm using gradio inside a docker that has the pytorch3d environment all set up.

johko:
Doing a quick search for "dockerfile gradio" with the new full-text search also reveals some spaces, which seem to be using Gradio from a docker container: https://huggingface.co/search/full-text?q=dockerfile+gradio&p=1&type=space
So it does not seem to be the rarest of use-cases I'd say

osanseviero:
Cool, thanks for explaining! Could you please solve the merge conflict? I can merge afterwards

johko:
thanks @osanseviero. the last commit has a confusing commit message, but should be good to go now

osanseviero (approved):
Thanks! 🔥

PR #680, opened 2023-02-15
WIP: add a sensitive content tag

julien-c:


allendorf (approved):

js/src/lib/interfaces/Types.ts
```diff
@@ -625,7 +625,7 @@ export const ALL_SUBTASK_TYPES = ALL_SUBTASKS.map(s => s.type);
 export const ALL_SUBTASK_TYPES_SET = new Set(ALL_SUBTASK_TYPES);
 
 
-export const TAG_SENSITIVE_CONTENT = "sensitive";
+export const TAG_NFAE_CONTENT = "nfae";
```
julien-c, line 628: hmm this is not super discoverable/explicit naming, no?

allendorf: I can rename it to `not-for-all-eyes` but I think the ethics folks are set on "not for all eyes" as the term they prefer instead of sensitive for now

allendorf: https://github.com/huggingface/hub-docs/pull/704

julien-c: yes i think `not-for-all-eyes` is better in that case

maybe underscores instead of dashes? seems more consistent with the existing "special" tags:
image

Pierrci: lol I was about to comment in #704 to have dashes instead, but why not XD

PR #682, opened 2023-02-16
🚩 Remove mentions to webhooks-explorers

coyotte508:


julien-c (commented):
other option was to replace the tip by a launch one: 

Webhooks are now publicly available!

coyotte508:
> other option was to replace the tip by a launch one:
> 
> Webhooks are now publicly available!

Is that a request? x)

julien-c:
no just a question/suggestion:)

julien-c (approved):

PR #690, opened 2023-02-22
Minor nit on docker Space examples

merveenoyan:
@osanseviero


osanseviero (approved):
Thanks

julien-c (changes requested):
can we wait a little bit before merging this one? Also i think we should split VSCode into its own line

PR #702, opened 2023-03-01
Added description to image-to-text

merveenoyan:
@osanseviero

### DIFF
--- a/tasks/src/image-to-text/data.ts
+++ b/tasks/src/image-to-text/data.ts
@@ -67,7 +67,7 @@ const taskData: TaskDataCustom = {
 			id:          "johko/capdec-image-captioning",
 		},
 	],
-	summary:      "",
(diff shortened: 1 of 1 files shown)


tasks/src/image-to-text/data.ts
```diff
@@ -67,7 +67,7 @@ const taskData: TaskDataCustom = {
 			id:          "johko/capdec-image-captioning",
 		},
 	],
-	summary:      "",
+	summary:      "Image to text models output a text from a given input. Image captioning or optical character recognition can be considered as most common applications of image to text.",
```
osanseviero, line 70:
```suggestion
	summary:      "Image to text models output a text from a given image. Image captioning or optical character recognition can be considered as the most common applications of image to text.",
```

osanseviero (approved):
Thanks!

PR #704, opened 2023-03-06
Rename nfae -> not-for-all-eyes 

allendorf:


js/src/lib/interfaces/Types.ts
```diff
@@ -649,7 +649,6 @@ export const OTHER_TAGS_SUGGESTIONS = [
 	"code",
 	"climate",
 	"medical",
-	TAG_NFAE_CONTENT,
```
Pierrci, line 652: We should keep it here (this way, it will be suggested in the metadata UI when filling a new model/dataset card, which is something we want, I think)

Removing it from the "Others" tags list in `/datasets` should be done in moon-landing directly if that's why you were removing it from here.

Pierrci (approved):

Issue #712, opened 2023-03-20
Tracking integration for text-to-video


osanseviero:
## Tracking integration of task - Text to Video

Note that you're not expected to do all of the following steps. This PR helps track all the steps required to get a new task fully supported in the Hub 🔥

- [ ] Integration with Inference API. Select at least one of the following:
  - [ ] Added a `transformers` `pipeline`
  - [ ] Added to Community Inference API for 3rd party library
  - [ ] Added to Community Inference API for generic
- [ ] Added basic UI elements (icon, order specification, etc)
- [ ] Added a widget

Integration guide: https://hf.co/docs/hub/models-tasks

osanseviero:
Example model https://huggingface.co/damo-vilab/modelscope-damo-text-to-video-synthesis
Others already tagged

osanseviero:
We should start with the UI elements (icon, order specification, etc) as per https://huggingface.co/docs/hub/models-tasks#ui-elements

gary149:
Awesome task! @Reeray is working on the icon fyi.

Reeray:
Here is the text-to-video icon. WDYT?
![text-to-video icon](https://user-images.githubusercontent.com/68782870/226711619-33afaf4f-a7f4-4919-a313-f5c2eeeaf756.svg)

Reeray:
![text-to-video icon](https://user-images.githubusercontent.com/68782870/226945854-6bcb3cd1-9a1e-4b04-8738-ddbc1ad4de2d.svg)

PR #714, opened 2023-03-22
Add text-to-video filter

osanseviero:
#712


js/src/lib/interfaces/Types.ts
```diff
@@ -578,6 +578,11 @@ export const PIPELINE_DATA = ensureRecordOfPipelines({
 		color:        "blue",
 		hideInModels: true,
 	},
+	"text-to-video": {
+		name:     "Text to Video",
```
gary149, line 582: image
Maybe Text-to-Video to be consistent with Text-to-Image, or the opposite?

julien-c: yes i would add dashes too

PR #716, opened 2023-03-23
Fix #714

Pierrci:
@osanseviero it wasn't caught because the CI only checks the files in `js/src/lib/interfaces`, and the faulty file is in `tasks/src`


Pierrci:
Ok fixed directly by 88e5c61374804be870e4c8d7106322b0b38a12f4 it seems, closing this :)

PR #727, opened 2023-03-30
refactor updateUrl to remove trailing ? character if no parameters

severo:
Now, if the list of query parameters is emptied, no trailing `?` is present.


osanseviero:
Hi there! This part of the code now lives in huggingface.js, so I'll close this PR. If it's still valid, feel free to open a PR in that repo. Thanks!

severo:
hmm, I don't remember the use case where I got the trailing `?`. Nevermind, it's not very important

PR #742, opened 2023-04-12
Update JAX GPU instructions on Spaces

pcuenca:
`cuda11_pip` is new and it resolves to the appropriate wheel better than the old method.


osanseviero (approved):
Thanks!

PR #745, opened 2023-04-13
Fix broken bold face

meg-huggingface:
Minor. Fixing the "Section Overview" that isn't appropriately bolded under "Uses".


osanseviero (approved):
Thanks!

PR #754, opened 2023-04-22
Add base_path config to spaces

coyotte508:
Cc @Pierrci @julien-c 

Thinking about it maybe we should rename the param to `initial_url` before making it widely available?

(cc @christophe-rannou for visibility, although this config is purely on the hub-side)


julien-c:
no idea 🤷 

i think the current name is ok, it's gonna be the most frequent use case no?

coyotte508:
ok then it's ready to be documented / merged

julien-c (approved):
lgtm then:)

PR #761, opened 2023-05-02
[Docs] Resolve URL linking to wrong library tag

tomaarsen:
Hello!

## Pull Request overiew
* Replace the library tag `span_marker` with the correct `span-marker` in the SpanMarker docs.

## Details
This originates from before I changed from `span_marker` to `span-marker`, I just missed this one case.

Before: https://huggingface.co/models?library=span_marker
After: https://huggingface.co/models?library=span-marker

- Tom Aarsen

### DIFF
--- a/docs/hub/span_marker.md
+++ b/docs/hub/span_marker.md
@@ -4,7 +4,7 @@
 
 ## Exploring SpanMarker in the Hub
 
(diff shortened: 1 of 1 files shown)


tomaarsen:
All URLs from the SpanMarker docs page work as intended now.

osanseviero (approved):
Thanks!

PR #765, opened 2023-05-03
Update img2img request

mishig25:
Widget counterpart for https://github.com/huggingface/api-inference-community/pull/244

### DIFF
--- a/js/src/lib/components/InferenceWidget/widgets/ImageToImageWidget/ImageToImageWidget.svelte
+++ b/js/src/lib/components/InferenceWidget/widgets/ImageToImageWidget/ImageToImageWidget.svelte
@@ -94,9 +94,9 @@
 		}
 
 		const requestBody = {
-			image: imageBase64,
+			inputs: imageBase64,
 			parameters: {
-				inputs: trimmedPrompt,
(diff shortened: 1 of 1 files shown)


js/src/lib/components/InferenceWidget/widgets/ImageToImageWidget/ImageToImageWidget.svelte
```diff
@@ -94,9 +94,9 @@
 		}
 
 		const requestBody = {
-			image: imageBase64,
+			inputs: imageBase64,
```
radames, line 97: hi @mishig25, considering last review from @Narsil we might not need to rename this inputs to image, actually both should work. However the parameters, prompts are correctly changed!

Narsil: Either way works for me.
`inputs` is more standardized across the board so I'm in favor of the change.

The API however must accept both (for backward compatibility)

radames:
the prompt parameter is already on prod, so this can be merge with either change for (inputs->image or image->inputs), ps still depends on https://github.com/huggingface/api-inference-community/pull/246

PR #769, opened 2023-05-04
add bigcode license

lvwerra:
Adds the BigCode OpenRAIL license to the Hub.


julien-c (commented):
it's a specific license? or we can use `openrail` maybe?

osanseviero (approved):

osanseviero:
It's a specific license that was changed for BigCode

osanseviero:
Internal discussion: https://huggingface.slack.com/archives/C03M3LKL4C8/p1683016451849279?thread_ts=1682626465.808009&cid=C03M3LKL4C8

julien-c:
yes but my general point is that we should use the same "license tag" for multiple models otherwise the license tag makes no sense

julien-c:
but ok to add this one, it's a more general point

PR #770, opened 2023-05-04
Update security.md

Michellehbn:
Adding a quick general snippet on GDPR


julien-c (approved):

julien-c:
nice, let's link customers to there now

GG @Michellehbn

PR #793, opened 2023-05-21
Update data.ts

balugans:
Added one dataset for depth estimation from existing datasets in HF.

### DIFF
--- a/tasks/src/depth-estimation/data.ts
+++ b/tasks/src/depth-estimation/data.ts
@@ -3,8 +3,8 @@ import type { TaskDataCustom } from "../Types";
 const taskData: TaskDataCustom = {
 	datasets: [
 		{
-			decription: "NYU Depth Dataset"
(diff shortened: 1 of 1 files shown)


tasks/src/depth-estimation/data.ts
```diff
@@ -1,7 +1,12 @@
 import type { TaskDataCustom } from "../Types";
 
 const taskData: TaskDataCustom = {
-	datasets: [],
+	datasets: [
+		{
+			decription: "NYU Depth Dataset"
```
merveenoyan, line 6:
```suggestion
			description: "NYU Depth Dataset",
```

merveenoyan: I feel like it would be better to add a simpler explanation to description, WDYT?

balugans: Thanks! I updated the corrections. In addition, I added some more explanation for the dataset.

PR #795, opened 2023-05-23
Added subtasks to `zero-shot-classification`

cdreetz:
Added subtasks

### DIFF
--- a/tasks/src/zero-shot-classification/about.md
+++ b/tasks/src/zero-shot-classification/about.md
@@ -33,6 +33,10 @@ pipe("I have a problem with my iphone that needs to be resolved asap!",
 >>> {'sequence': 'I have a problem with my iphone that needs to be resolved asap!!', 'labels': ['urgent', 'phone', 'computer', 'not urgent', 'tablet'], 'scores': [0.504, 0.479, 0.013, 0.003, 0.002]}
 ```
 
+## Subtasks / Use cases
+The general task of zero shot classification has a few subtasks that fall within it for different cases like image and language.  In natural language one subtasks is that as used above of sentiment analysis, cases where a model wasn't necessarily trained on what it is being used for, but through what it has learned during training and what it can infer through similarity, it can be applied to new cases of analyzing sentiment.  During natural language conversation, a model not have been directly trained on a concept, but with what it was trained on and its ability to infer through similarities it can have conversations on new concepts.  
+
+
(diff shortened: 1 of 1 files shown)


tasks/src/zero-shot-classification/about.md
```diff
@@ -33,6 +33,10 @@ pipe("I have a problem with my iphone that needs to be resolved asap!",
 >>> {'sequence': 'I have a problem with my iphone that needs to be resolved asap!!', 'labels': ['urgent', 'phone', 'computer', 'not urgent', 'tablet'], 'scores': [0.504, 0.479, 0.013, 0.003, 0.002]}
 ```
 
+## Subtasks / Use cases
+The general task of zero shot classification has a few subtasks that fall within it for different cases like image and language.  In natural language one subtasks is that as used above of sentiment analysis, cases where a model wasn't necessarily trained on what it is being used for, but through what it has learned during training and what it can infer through similarity, it can be applied to new cases of analyzing sentiment.  During natural language conversation, a model not have been directly trained on a concept, but with what it was trained on and its ability to infer through similarities it can have conversations on new concepts.
```
merveenoyan, line 37:
```suggestion
In natural language one subtasks is that as used above of sentiment analysis, cases where a model wasn't necessarily trained on what it is being used for, but through what it has learned during training and what it can infer through similarity, it can be applied to new cases of analyzing sentiment. 
```
This sentence is a bit too long to read, is it okay if we make it shorter? 😊

merveenoyan: BTW in HF there's two separate zero shot tasks: `zero-shot-classification` is for NLP and there's a separate `zero-shot-image-classification`.

merveenoyan (commented):
Thanks a lot for tackling this!

tasks/src/zero-shot-classification/about.md
merveenoyan: I opened it up internally :) is it ok if we could stick to `zero-shot-classification` for now?

merveenoyan:
Hey @cdreetz do you need any help with this?

merveenoyan:
Hey @cdreetz,
We have moved the codes for Tasks project to [huggingface.js repository](https://github.com/huggingface/huggingface.js) so if you feel like continuing this pull request you can move your work over there. Many thanks! 🤗

Issue #792, opened 2023-05-20
Completion of existing tasks  🪢❤️  


merveenoyan:
This issue is to complete or improve existing task pages: [hf.co/tasks](https://huggingface.co/tasks/). These pages are wikipedia of machine learning, and are crowdsourced. For each page, you could add an example Space or a more cutting edge model too. You can also add anything you think is useful information about the task to the use cases, task variants and so on. 

On how to start contributing, you can refer to #369 (it has all the information you need to start contributing!) 😊 

Depth Estimation
- [x] Example dataset
- [ ] Metrics and explanation

Image to Image
- [ ] Example dataset
- [ ] Useful resources

Object detection
- [ ] Example datasets

Video classification
- [ ] Example dataset

Zero-shot Image Classification
- [ ] Example dataset

Table Question Answering
- [ ] Task Variants
- [ ] Useful resources

Audio-to-Audio
- [x] Example dataset

Text-to-image
- [ ] Metrics
- [ ] Inference snippet
- [ ] Task variants

Unconditional Image Generation
- [ ] Inference snippet 

Zero-shot Classification
- [ ] Use Cases
- [ ] Metrics
- [x] Most used datasets
- [ ] Example Spaces
- [ ] Subtasks

Reinforcement Learning
- [x] Task Variants
- [ ] Model for the widget
- [x] Small snippet for the task

Feature extraction
- [ ] Metrics
- [ ] Use cases (needs improvement)
- [ ] Example Space

kishore-s-15:
@merveenoyan I would love to contribute to Zero-shot Image Classification and Table Question Answering. Could you kindly assign me for those tasks.

balugans:
@merveenoyan I would love to handle depth estimation.

jayant-yadav:
@merveenoyan I would like to contribute for Text-to-Image. Please assign me that task :)

samedovzaur:
would love to contribute @merveenoyan  :)

snehilsanyal:
Hey @merveenoyan I would like to contribute to Object Detection Task and Image-to-Image task.
For Object Detection I am planning to include a newer model Yolo-NAS as an sample example, will also search for newer datasets. 
For Image-to-Image task I would search for datasets and resources and try to include them. 
Please assign me these tasks. 
PS: First timer in contributing to hub-docs :D

younver:
Hey @merveenoyan, I'd like to contribute. Seems like audio-to-audio task left so I can try my best for that.

[...]

PR #807, opened 2023-05-29
Added `object-detection` dataset

znym:
Added a dataset for isssue #792

### DIFF
--- a/tasks/src/object-detection/data.ts
+++ b/tasks/src/object-detection/data.ts
@@ -1,13 +1,17 @@
 import type { TaskDataCustom } from "../Types";
 
 const taskData: TaskDataCustom = {
-	datasets: [
-		{
-			// TODO write proper description
-			description: "Widely used benchmark dataset for multiple Vision tasks.",
-			id:          "merve/coco2017",
-		},
+		datasets: [
+	  {
+		description: "Widely used benchmark dataset for multiple Vision tasks.",
+		id: "merve/coco2017",
+	  },
+	  {
+		description: "Car Object Detection dataset with annotated bounding boxes around cars.",
+		id: "nihany/car-object-detection",
+	  },
 	],
+
 	demo: {
 		inputs: [
 			{


tasks/src/object-detection/data.ts
```diff
@@ -1,13 +1,17 @@
...
-	datasets: [
-		{
-			// TODO write proper description
-			description: "Widely used benchmark dataset for multiple Vision tasks.",
-			id:          "merve/coco2017",
-		},
+		datasets: [
+	  {
+		description: "Widely used benchmark dataset for multiple Vision tasks.",
+		id: "merve/coco2017",
```
merveenoyan, line 7: I think there are many generic object detection datasets on Hub compared to this :)

tasks/src/object-detection/data.ts
```diff
@@ -1,13 +1,17 @@
...
+	  },
+	  {
+		description: "Car Object Detection dataset with annotated bounding boxes around cars.",
```
merveenoyan, line 10: this one sounds good!

merveenoyan (commented):
Thanks a lot for this PR! @znym

tasks/src/object-detection/data.ts
znym: Hi @merveenoyan, thank you so much for the review. Could you please clarify me please if I need to make any changes or what would be the next step I need to follow.

merveenoyan: I think there's only above review that you can edit and nothing else for now :)

kadirnar:
@merveenoyan , @znym 

Could you take a look at this for the Coco dataset?

https://huggingface.co/datasets/detection-datasets/coco

merveenoyan:
Hey @znym,
We have moved the codes for Tasks project to [huggingface.js repository](https://github.com/huggingface/huggingface.js) so if you feel like continuing this pull request you can move your work over there. Many thanks! 🤗

PR #808, opened 2023-05-30
Update tasks/reinforcement-learning/about.md

rickyliang:
De-dup links in Useful Resources and fix broken Notebooks links.
Notebook links prefer linking directly to github rather than colab for optionality.


merveenoyan (approved):
Thanks a lot for this! Very neat fix ♥️

PR #812, opened 2023-05-31
add bertopic task mapping

davanstrien:
cc @osanseviero, this adds a missing task to library mapping for BERTopic


osanseviero (approved):
Thanks!

PR #818, opened 2023-06-01
Soc2 and other small things

julien-c:


julien-c:
merging as no reviewer showed up =)

julien-c:
cc @jagwar for viz

PR #828, opened 2023-06-06
Update repositories-getting-started.md

severo:
see https://github.com/huggingface/moon-landing/issues/2507 (internal) as a reference

### DIFF
--- a/docs/hub/repositories-getting-started.md
+++ b/docs/hub/repositories-getting-started.md
@@ -133,7 +133,7 @@ And you're done! You can check your repository on Hugging Face with all the rece
 </div>
 
 <Tip>
-You might find it useful to setup a [git credential helper](https://git-scm.com/docs/gitcredentials) to avoid filling your username and password on every push operation.
+If you cloned the repository with HTTP, you might be asked to fill your username and password on every push operation. The simplest way to avoid repetition is to [switch to SSH](#cloning-repositories), instead of HTTP. Alternatively, if you have to use HTTP, you might find it helpful to setup a [git credential helper](https://git-scm.com/docs/gitcredentials#_avoiding_repetition) to autofill your username and password.
 </Tip>
 
 ## Viewing a repo's history
--- a/docs/hub/repositories-getting-started.md
+++ b/docs/hub/repositories-getting-started.md
@@ -133,7 +133,9 @@ And you're done! You can check your repository on Hugging Face with all the rece
 </div>
 
 <Tip>
+
 If you cloned the repository with HTTP, you might be asked to fill your username and password on every push operation. The simplest way to avoid repetition is to [switch to SSH](#cloning-repositories), instead of HTTP. Alternatively, if you have to use HTTP, you might find it helpful to setup a [git credential helper](https://git-scm.com/docs/gitcredentials#_avoiding_repetition) to autofill your username and password.
+
 </Tip>
 
 ## Viewing a repo's history
--- a/docs/hub/repositories-getting-started.md
+++ b/docs/hub/repositories-getting-started.md
@@ -132,6 +132,9 @@ And you're done! You can check your repository on Hugging Face with all the rece
 <img class="hidden dark:block" src="https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/hub/repo_with_files-dark.png"/>
 </div>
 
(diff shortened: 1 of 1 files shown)


docs/hub/repositories-getting-started.md
```diff
@@ -132,6 +132,9 @@ And you're done! You can check your repository on Hugging Face with all the rece
 <img class="hidden dark:block" src="https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/hub/repo_with_files-dark.png"/>
 </div>
 
+<Tip>
+You might find it useful to setup a [git credential helper](https://git-scm.com/docs/gitcredentials) to avoid filling your username and password on every push operation.
```
julien-c, line 136: i know its weird but i think GitHub's doc about this is better than https://git-scm.com/docs/gitcredentials

Or maybe better, do a short version of GitHub's doc in our doc

severo: hmmm, the only thing I find seems to be specific to GitHub: https://docs.github.com/en/get-started/getting-started-with-git/caching-your-github-credentials-in-git. Maybe they changed recently (_Password-based authentication for Git has been removed in favor of more secure authentication methods_)

julien-c: then maybe a super short extract from the git doc

BTW you also need to mention and link to the git+ssh doc, cc @SBrandeis (it's probably the simplest for anyone having trouble with setting up git-credential)

severo: I changed a bit, hopefully it's enough (super short extract from the git doc -> I think we should not push that solution, and it depends on every user OS / configuration, so... better to let them figure how to set it up)

alvarobartt:
Hi @severo, I also had another suggestion regarding the helper content showed when creating a new `Pull request` in the HuggingFace Hub, as you can see in the screenshot below:

image

My suggestion is: could we also include there the copy-paste command to clone the repository via SSH? I guess that file is not publicly accessible (if so, I couldn't find it), so maybe you can escalate this request to however applies if you think it's worth including!

Thanks 🤗

docs/hub/repositories-getting-started.md
```diff
@@ -132,6 +132,9 @@ And you're done! You can check your repository on Hugging Face with all the rece
 <img class="hidden dark:block" src="https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/hub/repo_with_files-dark.png"/>
 </div>
 
+<Tip>
+If you cloned the repository with HTTP, you might be asked to fill your username and password on every push operation. The simplest way to avoid repetition is to [switch to SSH](#cloning-repositories), instead of HTTP. Alternatively, if you have to use HTTP, you might find it helpful to setup a [git credential helper](https://git-scm.com/docs/gitcredentials#_avoiding_repetition) to autofill your username and password.
```
alvarobartt, line 136: LGMT, but shouldn't we point users to `Git over SSH` instead? See https://huggingface.co/docs/hub/security-git-ssh

```suggestion
If you cloned the repository with HTTP, you might be asked to fill your username and password on every push operation. The simplest way to avoid repetition is [clone it via SSH](#cloning-repositories), instead of HTTP. If SSH is not configured yet, please refer to [Git over SSH](./security-git-ssh). Alternatively, if you have to use HTTP, you might find it helpful to setup a [git credential helper](https://git-scm.com/docs/gitcredentials#_avoiding_repetition) to autofill your username and password.
```

docs/hub/repositories-getting-started.md
```diff
@@ -132,6 +132,9 @@ And you're done! You can check your repository on Hugging Face with all the rece
...
+</Tip>
```
stevhliu, line 137: Need to add a blank line before and after `<Tip>` so the links inside it are properly rendered!

```suggestion
<Tip>

If you cloned the repository with HTTP, you might be asked to fill your username and password on every push operation. The simplest way to avoid repetition is to [switch to SSH](#cloning-repositories), instead of HTTP. Alternatively, if you have to use HTTP, you might find it helpful to setup a [git credential helper](https://git-scm.com/docs/gitcredentials#_avoiding_repetition) to autofill your username and password.

</Tip>
```

stevhliu (approved):
The content looks good to me!

PR #836, opened 2023-06-08
Update ML-Agents documentation

simoninithomas:
# Context
- We finally integrated Hugging Face on the official ML-Agents repository. Therefore we need to update our documentation (setup process) and solve some markdown visual bugs).

### DIFF
--- a/docs/hub/ml-agents.md
+++ b/docs/hub/ml-agents.md
@@ -17,11 +17,12 @@ All models on the Hub come up with useful features:
 </div>
 
 ## Install the library
+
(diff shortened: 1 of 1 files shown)


simoninithomas:
@mishig25 do you know why there's a problem with `` markdown? It's after the <div> html tag so maybe it's because of that but I'm not sure 🤔

osanseviero (approved):
I don't see any issue in https://moon-ci-docs.huggingface.co/docs/hub/pr_836/en/ml-agents . I think we can merge this.

Thanks for the update!

PR #840, opened 2023-06-12
Add `lzh` to Language.ts

mariosasko:
"Resolves" https://github.com/huggingface/datasets/issues/5943.

### DIFF
--- a/js/src/lib/interfaces/Language.ts
+++ b/js/src/lib/interfaces/Language.ts
@@ -1238,5 +1238,9 @@ export const LANGUAGES: Record<string, Language> = {
 		name: "Erzya",
 		nativeName: "эрзянь кель",
 	},
+	lzh: {
+		code: "lzh",
+		name: "Literary Chinese",
+	},
 };
 


js/src/lib/interfaces/Language.ts
```diff
@@ -1238,5 +1238,9 @@ export const LANGUAGES: Record<string, Language> = {
 		name: "Erzya",
 		nativeName: "эрзянь кель",
 	},
+	lzh: {
+		code: "lzh",
+		name: "Literary Chinese",
```
julien-c, line 1243:
```suggestion
		name: "Literary Chinese",
		nativeName: "",
```

could you add a nativeName too, ie. how is Literary Chinese written in Literary Chinese?

ayaka14732: It should be "文言"

ayaka14732, line 1243:
```suggestion
		name: "Literary Chinese",
		nativeName: "文言",
```

mariosasko: Thanks!

mariosasko: Yes, based on the Wikipedia page and some Literary Chinese resources, this seems correct (the Wikipedia page lists two native names, but one represents Classical Chinese, which is not identical to Literary Chinese).

ayaka14732: I am Chinese and I can confirm that "文言" is correct.

"Literary Chinese" and "Classical Chinese" are both correct and common translation of "文言". Since "Literary Chinese" is the translation used in https://github.com/huggingface/datasets/blob/7fcbe5b1575c8d162b65b9397b3dfda995a4e048/src/datasets/utils/resources/languages.json#L3880, we can just follow the translation.

---

Long explaination based on my understanding (may be inaccurate):

The term "文言" (**Modern Mandarin**: Wényán; **Modern Cantonese**: man4 jin4; **Early Middle Chinese**: mun-ngyon; **literally**: Literary Language) is often translated as both "Literary Chinese" and "Classical Chinese". Both translations are very commonly used and they highlight different aspects of 文言's nature and historical context.

"Literary Chinese" emphasizes the "literary" aspect of 文言. For centuries, 文言 was the standard form of written Chinese, used in official documents, literature, and scholarly works. Despite the vast changes in spoken Chinese over these centuries, 文言 remained the primary choice for written communication and expression until the early 20th century, showcasing its profound literary characteristics.

"Classical Chinese" emphasizes the "classical" aspect of 文言. 文言 originated from the language spoken during the Spring and Autumn and the Warring States periods (770-221 BC) of ancient China. This term underscores its antiquity and the fact that it's the traditional, time-honoured form of the Chinese language, which laid the foundation for later forms of Chinese.

The use of either "Literary Chinese" or "Classical Chinese" thus depends on the context and the specific aspect of 文言 one wishes to highlight.

---

> The Wikipedia page lists two native names, but one represents Classical Chinese, which is not identical to Literary Chinese.

For the name in Literary Chinese, "文言" (literally "Literary Language") is the most common one, and "古典漢語" (the one corresponds to "Classical Chinese") is almost never used.

---

The word "文言" is still very commonly used in modern China, and it is a compulsory subject in high school. Below is a question from the Classical Chinese section of the 2023 college entrance examination (Gaokao) that took place just a week ago.

mariosasko: Thanks for the detailed explanation!

osanseviero (approved):
Thanks! Feel free to merge

PR #853, opened 2023-06-16
Added new models to tasks

merveenoyan:


tasks/src/text-generation/data.ts
```diff
@@ -101,7 +105,7 @@ const taskData: TaskDataCustom = {
 		},
 	],
 	summary:      "Generating text is the task of producing new text. These models can, for example, fill in incomplete text or paraphrase.",
-	widgetModels: ["gpt2"],
+	widgetModels: ["bigscience/bloom-560m"],
```
osanseviero, line 108: I think we can use falcon here

osanseviero (approved):
Thanks

tasks/src/text-generation/data.ts
merveenoyan: I used instruction fine-tuned 7B version as for the other ones widget was turned off (it was reason why I didn't pick it as well :/)

osanseviero (approved):
Thanks!

julien-c:
hope to move away from gpt2 on https://huggingface.co/tasks/text-generation soon, it feels a bit 2019 on HF for now :)

PR #863, opened 2023-06-23
Add JS snippets to task pages

merveenoyan:
This PR adds `huggingface.js` snippets for task inferences. I wasn't super sure of the snippets that take multiple inputs so this one only adds the ones that have single input for now. Will add them later.

### DIFF
--- a/tasks/src/audio-classification/about.md
+++ b/tasks/src/audio-classification/about.md
@@ -48,6 +48,18 @@ data = query("sample1.flac")
 # {'label': 'sad', 'score': 0.07}]
 ```
 
+You can use [huggingface.js](https://github.com/huggingface/huggingface.js) to infer with audio classification models on Hugging Face Hub.
+
+```javascript
+import { HfInference } from "@huggingface/inference";
+
+const inference = new HfInference(HF_ACCESS_TOKEN);
+await inference.audioClassification({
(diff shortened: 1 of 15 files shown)


coyotte508 (approved):
Thanks, LGTM!

xenova (commented):
Looks great! 🥳

Just a general comment about the code snippets: when I was testing out the HF inference API a while ago, I found it a bit annoying to copy+paste the snippets, run the code, and then see nothing appear. Of course, the reason was because the result was being awaited but not saved to a variable. So, I'd write `const output = ... ; console.log(output);` to see the output. I know this is how it's done across the huggingface.js documentation, so it's probably not a big thing to change. Perhaps an alternative is just to show what the output will look like, with a "show output" `<summary>` tag (or something similar?)

Thoughts? 😋

merveenoyan:
@xenova we are trying to keep these as minimal as possible, but I'd like to have @osanseviero's comment :)

xenova:
> @xenova we are trying to keep these as minimal as possible, but I'd like to have @osanseviero's comment :)

Gotcha! 😄 I see that some of the pages (e.g. [object-detection](https://huggingface.co/tasks/object-detection)) already include example output:
image

so, it might not be necessary :)

Issue #735, opened 2023-04-05
Add a code snippet for image-to-text


osanseviero:
Example repo: https://huggingface.co/Salesforce/blip-image-captioning-base 

We should also add it in https://huggingface.co/docs/api-inference/detailed_parameters

regstuff:
If it's not too much trouble, I'd suggest adding an example for each of the tasks here: https://huggingface.co/models
Right now I'd say bout 30% of the tasks are not detailed in the api-inference page.
HF ecosystem is pretty diverse and every task seems to have different params to consider. Examples in the API would be very helpful.
Thanks

osanseviero:
Same for document understanding

key88sf:
@osanseviero Also some details about how to deploy these image-to-text inference models on AWS Sagemaker (e.g. what "Task" do we select) ?

KindaCallam-io:
Hey is there any update on this? 
Has anybody been able to get this working and potentially provide me with an example?

xianbaoqian:
Thanks for fixing the PR! Verified that this config is working. CodeSpace can create pull request to hub-docs and run the server. We can close it now.

Proof: https://github.com/huggingface/hub-docs/pull/876

PR #876, opened 2023-07-03
add snippets for image-to-text

xianbaoqian:
image
image
image

Request: https://github.com/huggingface/hub-docs/issues/735

### DIFF
--- a/js/src/lib/inferenceSnippets/inputs.ts
+++ b/js/src/lib/inferenceSnippets/inputs.ts
@@ -67,6 +67,8 @@ const inputsFeatureExtraction = () =>
 
 const inputsImageClassification = () => `"cats.jpg"`;
 
+const inputsImageToText = () => `"cats.jpg"`;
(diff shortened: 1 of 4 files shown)


xianbaoqian:
(use gpt2 as an example)

coyotte508:
Do we want to promote the official libs ? cc @Wauplin 

For example:

```ts
import { HfInference } from "@huggingface/inference";

const inference = new HfInference("hf_XXXXXX");

const text = await inference.imageToText({
  data: await (await fetch('https://picsum.photos/300/300')).blob(),
  model: "gpt2"
});
```

osanseviero (approved):
LGTM, thanks!

xianbaoqian:
This new API looks cool.   Let me merge this one as it is. I'll create a new one to update all the snippets.   Maybe we can have both (if user would rather prefer not to use our library unfortunately).

PR #880, opened 2023-07-04
Add text-to-video spaces, resources, schema and more

merveenoyan:
Screenshot 2023-07-04 at 17 30 54


tasks/src/text-to-video/about.md
```diff
@@ -1,72 +1,38 @@
-Text-to-video models can generate videos from a given text prompt. They can be used in any application that requires generating temporally and spatially consistent sequence of images from text.  
-
-
 ## Use Cases
 
 ### Script-based Video Generation
-
 Text-to-video models can be used to create short-form video content from a provided text script. These models can be used to create engaging and informative marketing videos. For example, a company could use a text-to-video model to create a video that explains how their product works or that shows how their product can solve a customer's problem.
```
osanseviero, line 4:
```suggestion
Text-to-video models can be used to create short-form video content from a provided text script. These models can be used to create engaging and informative marketing videos. For example, a company could use a text-to-video model to create a video that explains how their product works.

```

tasks/src/text-to-video/about.md
```diff
@@ -1,72 +1,38 @@
...
-
-
 ## Use Cases
 
 ### Script-based Video Generation
-
 Text-to-video models can be used to create short-form video content from a provided text script. These models can be used to create engaging and informative marketing videos. For example, a company could use a text-to-video model to create a video that explains how their product works or that shows how their product can solve a customer's problem.
 
-
 ### Content format conversion
```
osanseviero, line 6: Nice examples, but I'm not sure we have a model that can generate videos with quality for these. We might have some assistive models, but nothing ready to do safety protocols/disaster relief

tasks/src/text-to-video/about.md
```diff
@@ -1,72 +1,38 @@
...
 
 
-
-
 ## Task Variants
-
 Text-to-video models have different variants based on inputs and outputs.
 
-
 ### Text-to-video Retrieval
```
osanseviero, line 19: Is this really using `text-to-video` models though? This is closer to semantic search/sentence similarity.

merveenoyan: I think you can use a text-to-video model in sort of feature extraction-y fashion and that's what they mention (sort of like semantic similarity)

tasks/src/text-to-video/about.md
merveenoyan: I agree

tasks/src/text-to-video/about.md
osanseviero: But this sounds more like a different task then (kinda like CLIP)

tasks/src/text-to-video/data.ts
```diff
@@ -44,12 +51,24 @@ const taskData: TaskDataCustom = {
...
+		{
+			description: "An application that generates video from text.",
+			id:          "fffiloni/zeroscope",
+		},
+		{
+			description: "An application that generates video from image and text.",
+			id:          "TempoFunk/makeavid-sd-jax",
+		},
+	],
+	summary:       "Text-to-video is the task of generating videos from text. Text-to-video models can be used in any application that requires generating temporally and spatially consistent sequence of images from text. ",
```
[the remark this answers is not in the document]
merveenoyan: Removing

PR #883, opened 2023-07-05
Use model's pipeline_tag not the transformers metadata's one

julien-c:
so that we honor a model card overridden pipeline_tag for instance


osanseviero (approved):
Sounds good, thanks!

PR #884, opened 2023-07-06
Document data encoding / decoding in inference.md

timsteuer:
Custom endpoints need to encode and decode the data passed to the callback functions. 

This is a little unclear in the current documentation and one may assumes the data has already `dict` format (like I did). Thus, I updated the examples to make the encoding and decoding of the JSON string arguments explicit.


philschmid (approved):
Thank you for improving the docs.

PR #889, opened 2023-07-10
Add `credentials` to model status fetch

mishig25:
Just like the fetch to infrence, fetch to model status should have `credentials` attribute set as well

Had to change bunch of files but the only change that needs to be reviewed is https://github.com/huggingface/hub-docs/pull/889/files#diff-db7f2d2a8101371aad01e3f7679a486c69e58a9082807f80aa04697ab1471f9c


julien-c:
thanks for fixing

osanseviero (commented):
Thank you! :fire:

PR #891, opened 2023-07-11
Show Metadata UI instead of linking to old dataset tagging app

lhoestq:


julien-c (approved):
👍

BTW:
- if you ever want to add fields to the metadata GUI, should be easy to do so (a bit harder for complex data structures, but still possible – for instance @polinaeterna's metadata spec from a few weeks ago, etc)
- (tiny-nit) orthogonal to this PR, i would rename datasets-cards to dataset-cards (and redirect the old one to the new)

PR #895, opened 2023-07-16
Fix Inference API status check, allow modelLoadInfo to be passed as prop

aliabid94:
To reduce the number of calls made to the inference API status check, we let modeLoadInfo to be passed in as a prop - if not provided as a prop, it makes the api call directly as it always has. This allows the Model Page in moon-landing, which needs to make the status check request, reuse the model status instead of making another call. See discussion here: https://github.com/huggingface/moon-landing/pull/6975

Also I believe the status API got out of sync at some point with the actual API returned by api-inference - fixed now. This is why currently every model says `This model can be loaded on the Inference API on-demand` even when that isn't the case.

### DIFF
--- a/js/src/lib/components/InferenceWidget/shared/WidgetInfo/WidgetInfo.svelte
+++ b/js/src/lib/components/InferenceWidget/shared/WidgetInfo/WidgetInfo.svelte
@@ -6,7 +6,7 @@
 	export let model: WidgetProps["model"];
 	export let computeTime: string;
 	export let error: string;
-	export let modelLoadInfo: ModelLoadInfo;
+	export let modelLoadInfo: ModelLoadInfo | undefined;
 
 	const status = {
 		TooBig: "⚠️ This model is too large to be loaded by the inference API. ⚠️",
@@ -21,10 +21,13 @@
 	} as const;
 
 	function getStatusReport(
-		modelLoadInfo: ModelLoadInfo,
+		modelLoadInfo: ModelLoadInfo | undefined,
 		statuses: Record<LoadingStatus, string>,
 		isAzure = false
 	): string {
+		if (!modelLoadInfo) {
+			return "Unable to get model status.";
+		}
 		if (
 			modelLoadInfo.compute_type === "cpu" &&
 			modelLoadInfo.state === "Loaded" &&
@@ -36,6 +39,9 @@
 	}
 
 	function getComputeTypeMsg(): string {
+		if (!modelLoadInfo) {
+			return "Unknown compute type";
+		}
 		let compute_type = modelLoadInfo?.compute_type ?? "cpu";
 		if (compute_type === "cpu") {
 			return "Intel Xeon 3rd Gen Scalable cpu";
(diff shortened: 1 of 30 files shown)


js/src/lib/components/InferenceWidget/shared/helpers.ts
```diff
@@ -180,16 +180,14 @@ export async function getResponse<T>(
...
 	const output = await response.json();
-	if (response.ok && typeof output === 'object' && output.loaded !== undefined) {
-		const status = output.loaded ? 'loaded' : 'unknown';
-		const compute_type = output.compute_type;
-		return {status, compute_type}
+	if (response.ok) {
+		return output as ModelLoadInfo;
 	} else {
 		console.warn(response.status, output.error);
-		return {status: 'error'};
```
mishig25, line 192: @aliabid94 could you explain removal of this line? The error is handled somewhere?

js/src/lib/components/InferenceWidget/shared/types.ts
```diff
@@ -9,16 +9,19 @@ export interface WidgetProps {
...
+export type LoadingStatus = "Loaded" | "Loadable" | "TooBig";
 
 export type ComputeType = "cpu" | "gpu";
 
 export type ModelLoadInfo = {
-	status: LoadingStatus;
-	compute_type?: ComputeType;
+	loaded:       boolean;
+	state:        "Loaded" | "Loadable" | "TooBig";
+	compute_type: "cpu" | "ppu";
```
julien-c, line 23:
```suggestion
	compute_type: "cpu" | "gpu";
```
?

js/src/lib/components/InferenceWidget/shared/WidgetInfo/WidgetInfo.svelte
```diff
@@ -6,36 +6,42 @@
...
+		Loadable: "This model can be loaded on the Inference API on-demand.",
 	} as const;
 
 	const azureStatus = {
-		error: "⚠️ This model could not be loaded.",
-		loaded: "This model is loaded and running on AzureML Managed Endpoint",
-		unknown: "This model can be loaded loaded on AzureML Managed Endpoint",
+		TooBig: "⚠️ This model is too large to be loaded by the inference API.",
+		Loaded: "This model is loaded and running on AzureML Managed Endpoint",
+		Loadable: "This model can be loaded loaded on AzureML Managed Endpoint",
```
julien-c, line 20: are those the only possible states in the `api-inference` codebase?

js/src/lib/components/InferenceWidget/shared/helpers.ts
aliabid94: I'm using undefined to mean an unknown model status. either we get a ModelLoadInfo object if we communicated with the server or undefined if not.

aliabid94: e.g. here: https://github.com/huggingface/hub-docs/pull/895/files#diff-47bdcb4d33fde7883ff228c78a96272ed67aa281dfcbceef721e9a80fcb8a211R28

js/src/lib/components/InferenceWidget/shared/WidgetInfo/WidgetInfo.svelte
aliabid94: https://github.com/huggingface/api-inference/blob/1351b5c75e410fe3d5adfc315d50be4df4ee2a6d/master-rs/src/resources/status.rs#L75

osanseviero:
Hi there! This part of the code now lives in huggingface.js, so I'll close this PR.

PR #915, opened 2023-08-17
Support reinforcement learning and robotics as dataset tags

mariosasko:
Asked for in https://github.com/huggingface/datasets/issues/6125

### DIFF
--- a/js/src/lib/interfaces/Types.ts
+++ b/js/src/lib/interfaces/Types.ts
@@ -460,7 +460,6 @@ export const PIPELINE_DATA = ensureRecordOfPipelines({
 		name:           "Reinforcement Learning",
 		modality:       "rl",
 		color:          "red",
-		hideInDatasets: true,
 	},
 	"robotics": {
 		name:     "Robotics",
@@ -476,7 +475,6 @@ export const PIPELINE_DATA = ensureRecordOfPipelines({
 			},
 		],
 		color:          "blue",
-		hideInDatasets: true,
 	},
 	"tabular-classification": {
 		name:     "Tabular Classification",


lhoestq:
Sounds good to me :)

It would be empty so far though no ? We could add them and see if some datasets are coming up and if it stays empty we can re-hide them

mariosasko:
Yes, I had the same thought. Maybe @julien-c @osanseviero can share their opinions.

osanseviero:
There are some from @edbeeching for RL and 3 for robotics https://huggingface.co/datasets?other=robotics

I think adding a filter for 3 datasets is not super nice experience, so maybe just the RL ones for now? WDYT?

StoneT2000:
Just chiming in here. I think robotics is quite separate given there are many other approaches like imitation learning and behavior cloning that would not fall under RL. An RL tag and robotics tag would be great! And my lab will likely be uploading several models/datasets surrounding these soon for benchmarking robotic manipulation in simulation and working with sim2real transfer.

osanseviero (approved):
LGTM given the above :fire:

julien-c:
this has not been deployed yet, correct?

wanted to make a nice screenshot of the new tasks on https://huggingface.co/datasets =)

StoneT2000:
Looks like its up now. Thank you all for the fast work! Glad to be one of the first datasets tagged under robotics :)

osanseviero:
Awesome! @albertvillanova @simoninithomas let's do some comms for this :)

julien-c:
yaaaay, looks good @StoneT2000!! we're ready for more datasets!

image

PR #916, opened 2023-08-22
Make hub-docs compatible with HF Endpoints

beurkinger:
This PR contains modifications made so that `hf-endpoints` can import `hub-docs` as a Git submodule and use its components, utils and types (instead of having a local custom copy). This way the project can get every update pushed to `hub-docs` and we won't have to manually maintain dozens of files manually anymore. 

The main changes are:
- adding a `noModelLoading` props to Widgets. If set to false, it deactivates the API status check (which is not handled in this context of Endpoints).
- adding an `appendRepoPath` props to Widgets. If set to false, it prevents the fetching function from appending `/models/${repoId}` to the API url.
- fix some typing issues (typing is stricter on the Endpoints side)

This changes shouldn't affect moon-landing in any way, However, they probably are going to conflict with https://github.com/huggingface/hub-docs/pull/895 :'(

### DIFF
--- a/js/src/lib/components/InferenceWidget/shared/WidgetWrapper/WidgetWrapper.svelte
+++ b/js/src/lib/components/InferenceWidget/shared/WidgetWrapper/WidgetWrapper.svelte
@@ -1,77 +1,69 @@
 <script lang="ts">
-	import type { WidgetProps, ModelLoadInfo } from "../types";
-	import type { WidgetInputSample } from "../../../../interfaces/Types";
+import type { WidgetProps, ModelLoadInfo } from '../types';
+import type { WidgetInputSample } from '../../../../interfaces/Types';
 
-	import { onMount } from "svelte";
-	import IconCross from "../../../Icons/IconCross.svelte";
-	import WidgetInputSamples from "../WidgetInputSamples/WidgetInputSamples.svelte";
-	import WidgetInputSamplesGroup from "../WidgetInputSamplesGroup/WidgetInputSamplesGroup.svelte";
-	import WidgetFooter from "../WidgetFooter/WidgetFooter.svelte";
-	import WidgetHeader from "../WidgetHeader/WidgetHeader.svelte";
-	import WidgetInfo from "../WidgetInfo/WidgetInfo.svelte";
-	import WidgetModelLoading from "../WidgetModelLoading/WidgetModelLoading.svelte";
-	import { getModelLoadInfo } from "../../shared/helpers";
+import { onMount } from 'svelte';
+import IconCross from '../../../Icons/IconCross.svelte';
+import WidgetInputSamples from '../WidgetInputSamples/WidgetInputSamples.svelte';
+import WidgetInputSamplesGroup from '../WidgetInputSamplesGroup/WidgetInputSamplesGroup.svelte';
+import WidgetFooter from '../WidgetFooter/WidgetFooter.svelte';
+import WidgetHeader from '../WidgetHeader/WidgetHeader.svelte';
+import WidgetInfo from '../WidgetInfo/WidgetInfo.svelte';
+import WidgetModelLoading from '../WidgetModelLoading/WidgetModelLoading.svelte';
+import { getModelLoadInfo } from '../../shared/helpers';
 
-	export let apiUrl: string;
-	export let computeTime: string;
-	export let error: string;
-	export let isLoading = false;
-	export let model: WidgetProps["model"];
-	export let includeCredentials: WidgetProps["includeCredentials"];
-	export let modelLoading = {
-		isLoading: false,
-		estimatedTime: 0,
-	};
-	export let noTitle = false;
-	export let outputJson: string;
-	export let applyInputSample: (sample: Record<string, any>) => void =
-		({}) => {};
-	export let previewInputSample: (sample: Record<string, any>) => void =
-		({}) => {};
+export let apiUrl: string;
+export let computeTime: string;
+export let error: string;
+export let isLoading = false;
+export let model: WidgetProps['model'];
+export let modelLoading: {
+	isLoading: boolean;
+	estimatedTime: number;
+};
+export let noModelLoading = false;
+export let noTitle = false;
+export let outputJson: string;
+export let applyInputSample: (sample: Record<string, any>) => void = ({}) => {};
+export let previewInputSample: (sample: Record<string, any>) => void = ({}) => {};
 
-	let isMaximized = false;
-	let modelLoadInfo: ModelLoadInfo = { status: "unknown" };
-	let selectedInputGroup: string;
+let isMaximized = false;
+let modelLoadInfo: ModelLoadInfo = { status: true ? 'ignored' : 'unknown' };
+let selectedInputGroup: string;
 
-	const inputSamples: WidgetInputSample[] = (model?.widgetData ?? [])
-		.sort(
-			(sample1, sample2) =>
-				(sample2.example_title ? 1 : 0) - (sample1.example_title ? 1 : 0)
-		)
-		.map((sample, idx) => ({
-			example_title: `Example ${++idx}`,
-			group: "Group 1",
-			...sample,
-		}));
+const inputSamples: WidgetInputSample[] = (model?.widgetData ?? [])
+	.sort((sample1, sample2) => (sample2.example_title ? 1 : 0) - (sample1.example_title ? 1 : 0))
+	.map((sample, idx) => ({
+		example_title: `Example ${++idx}`,
+		group: 'Group 1',
+		...sample,
+	}));
 
-	const inputGroups: { group: string; inputSamples: WidgetInputSample[] }[] =
-		[];
-	for (const inputSample of inputSamples) {
-		const isExist = inputGroups.find(
-			({ group }) => group === inputSample.group
-		);
-		if (!isExist) {
-			inputGroups.push({ group: inputSample.group, inputSamples: [] });
-		}
-		inputGroups
-			.find(({ group }) => group === inputSample.group)
-			?.inputSamples.push(inputSample);
+const inputGroups: { group: string; inputSamples: WidgetInputSample[] }[] = [];
+for (const inputSample of inputSamples) {
+	const isExist = inputGroups.find(({ group }) => group === inputSample.group);
+	if (!isExist) {
+		inputGroups.push({ group: inputSample.group, inputSamples: [] });
 	}
+	inputGroups.find(({ group }) => group === inputSample.group)?.inputSamples.push(inputSample);
+}
 
-	$: selectedInputSamples =
-		inputGroups.length === 1
-			? inputGroups[0]
-			: inputGroups.find(({ group }) => group === selectedInputGroup);
+$: selectedInputSamples =
+	inputGroups.length === 1 ? inputGroups[0] : inputGroups.find(({ group }) => group === selectedInputGroup);
 
-	onMount(() => {
-		getModelLoadInfo(apiUrl, model.id, includeCredentials).then((info) => {
-			modelLoadInfo = info;
-		});
+onMount(() => {
+	if (noModelLoading) {
+		modelLoadInfo = { status: 'ignored' };
+		return;
+	}
+	getModelLoadInfo(apiUrl, model.id).then((info) => {
+		modelLoadInfo = info;
 	});
+});
 
-	function onClickMaximizeBtn() {
-		isMaximized = !isMaximized;
-	}
+function onClickMaximizeBtn() {
+	isMaximized = !isMaximized;
+}
 </script>
 
 <div
@@ -95,9 +87,7 @@
 					/>
 				{/if}
 				<WidgetInputSamples
-					classNames={!selectedInputSamples
-						? "opacity-50 pointer-events-none"
-						: ""}
+					classNames={!selectedInputSamples ? 'opacity-50 pointer-events-none' : ''}
 					{isLoading}
 					inputSamples={selectedInputSamples?.inputSamples ?? []}
 					{applyInputSample}
--- a/js/src/lib/components/InferenceWidget/shared/helpers.ts
+++ b/js/src/lib/components/InferenceWidget/shared/helpers.ts
@@ -64,6 +64,7 @@ async function callApi(
 	useCache = true,
 	includeCredentials = false,
 	isOnLoadCall = false,
+	appendRepoPath = true,
 ): Promise<Response> {	
 	const contentType = 'file' in requestBody && 'type' in requestBody['file']
 		? requestBody['file']['type']  
@@ -87,9 +88,9 @@ async function callApi(
 	const body: File | string = 'file' in requestBody
 		? requestBody.file
 		: JSON.stringify(requestBody);
-	
+
 	return await fetch(
-		`${url}/models/${repoId}`,
+		`${url}${appendRepoPath ? `/models/${repoId}` : ''}`,
 		{
 			method: "POST",
 			body,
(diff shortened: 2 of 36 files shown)


julien-c:
there are some style only changes (double quotes -> simple quotes etc) can we revert them to align to moon-landing coding prettier config?

beurkinger:
@julien-c weird I made sure there was no style change while working on it 🤔

js/src/lib/interfaces/Libraries.ts
```diff
@@ -73,7 +73,7 @@ export interface LibraryUiElement {
 
 function nameWithoutNamespace(modelId: string): string {
 	const splitted = modelId.split("/");
-	return splitted.length === 1 ? splitted[0] : splitted[1];
+	return splitted[1] ?? splitted[0] ?? "";
```
Pierrci, line 76:
```suggestion
	return splitted[1] ?? splitted[0];
```
I think that's enough, we'll necessarily have something in `splitted[0]` (at the minimum it will be `""`)

js/src/lib/components/PipelineTag/PipelineTag.svelte
```diff
@@ -1,14 +1,18 @@
...
 
-	export let classNames = "";
-	export let pipeline = "";
+import { PIPELINE_DATA } from '../../interfaces/Types';
+import PipelineIcon from '../PipelineIcon/PipelineIcon.svelte';
+
+export let classNames = '';
+export let pipeline = '';
+
+$: pipelineName = pipeline ? PIPELINE_DATA[<PipelineType>pipeline]?.name ?? pipeline : pipeline;
```
Pierrci, line 10: formatting needs to be reversed here too

js/src/lib/components/PipelineIcon/PipelineIcon.svelte
```diff
@@ -80,9 +80,11 @@
 		"text-to-video": IconTextToVideo,
 		"document-question-answering": IconDocumentQuestionAnswering,
 	};
+
+	$: component = pipeline in ICON_COMPONENTS ? ICON_COMPONENTS[<PipelineType>pipeline] : IconFillMask;
 </script>
 
 <svelte:component
-	this={ICON_COMPONENTS[pipeline] ?? IconFillMask}
+	this={component ?? IconFillMask}
```
Pierrci, line 88:
```suggestion
	this={component}
```

js/src/lib/components/InferenceWidget/shared/helpers.ts
```diff
@@ -180,8 +183,8 @@ export async function getResponse<T>(
 }
 
 
-export async function getModelLoadInfo(url: string, repoId: string, includeCredentials = false): Promise<ModelLoadInfo> {
-	const response = await fetch(`${url}/status/${repoId}`, {credentials: includeCredentials ? "include" : "same-origin"});
+export async function getModelLoadInfo(url: string, repoId: string): Promise<ModelLoadInfo> {
+	const response = await fetch(`${url}/status/${repoId}`);
```
Pierrci, line 187: are you undoing this change from #889 on purpose, and if so, are you sure about it? cc @mishig25

js/src/lib/components/InferenceWidget/shared/helpers.ts
```diff
@@ -87,9 +88,9 @@ async function callApi(
 	const body: File | string = 'file' in requestBody
 		? requestBody.file
 		: JSON.stringify(requestBody);
-	
+
 	return await fetch(
-		`${url}/models/${repoId}`,
+		`${url}${appendRepoPath ? `/models/${repoId}` : ''}`,
```
Pierrci, line 93: Not sure I understand here; is it because `url` will already include the model path in the case of IE?

js/src/lib/components/InferenceWidget/shared/helpers.ts
```diff
@@ -109,6 +110,7 @@ export async function getResponse<T>(
 	includeCredentials = false,
 	isOnLoadCall = false, // If true, the server will try to answer from cache and not do anything if not
 	useCache = true,
+	appendRepoPath = true,
```
Pierrci, line 113: we should refactor the whole thing into a `params` object, this is really too much now

js/src/lib/components/InferenceWidget/shared/helpers.ts
beurkinger: We don't have a "model path" in IE. Each endpoint only has a single API point (the `url` arg).

js/src/lib/components/PipelineIcon/PipelineIcon.svelte
beurkinger: Hmmm are you sure ?

js/src/lib/interfaces/Libraries.ts
beurkinger: yes but TypeScript doesn't know  about it :( 

Screen Shot 2023-08-25 at 2 12 31 PM

js/src/lib/components/InferenceWidget/shared/helpers.ts
beurkinger: It may be a merging issue, I'm going to look up.

js/src/lib/components/InferenceWidget/shared/helpers.ts
beurkinger: I agree. In another branch ?

js/src/lib/components/InferenceWidget/shared/helpers.ts
beurkinger: Fixed !

js/src/lib/components/InferenceWidget/widgets/FillMaskWidget/FillMaskWidget.svelte
```diff
@@ -109,7 +113,7 @@
 			// entries = body -> text-classificartion
 			// entries = body[0] -> summarization
 			const entries = (
-				model.pipeline_tag === "text-classification" ? body[0] ?? [] : body
+				body?.[0] && Array.isArray(body[0]) ? body[0] : body
```
beurkinger, line 116: Oh yeah, regarding this change: for some reason the data we get for "text-classification" is different between Endpoints and the Inference API. This change makes sure both response types are parsed correctly;

krampstudio:
> there are some style only changes (double quotes -> simple quotes etc) can we revert them to align to moon-landing coding prettier config?

If needed I've prepared https://github.com/huggingface/hub-docs/pull/924 bc I went into the same issue and we could fix the styles all at once

beurkinger:
@krampstudio I think I've fixed all styling issues in this PR. But thanks for setting up a prettier config, I think it will be useful :)

js/src/lib/components/PipelineTag/PipelineTag.svelte
beurkinger: Done !

julien-c:
can you rebase on top of main now that #924 has been merged?

julien-c:
Also we would need a review from @mishig25 to merge this (no big rush though as @Pierrci should also take a look)

js/src/lib/components/InferenceWidget/shared/WidgetWrapper/WidgetWrapper.svelte
```diff
@@ -67,8 +70,7 @@
 
 <div
 	class="flex w-full max-w-full flex-col
-	{isMaximized ? 'fixed inset-0 z-20 bg-white p-12' : ''}
-	{!modelLoadInfo ? 'hidden' : ''}"
+	{isMaximized ? 'fixed inset-0 z-20 bg-white p-12' : ''}"
```
Pierrci, line 73: this undoes something recently added in #935

^^ cc @julien-c @gary149, I think there's been some discussion about the disappearance of the widget while its state is loading, this would bring it back - @mishig25 are there any side-effects/downsides in not hiding the widget if we don't have the state yet?

(IIUC, with the other changes in this PR, `WidgetInfo` would still show nothing if no state information)

js/src/lib/components/InferenceWidget/widgets/TextGenerationWidget/TextGenerationWidget.svelte
```diff
@@ -128,7 +131,7 @@
 			if (output.length === 0) {
 				warning = "No text was generated";
 			} else if (model?.pipeline_tag === "text-generation") {
-				const outputWithoutInput = output.slice(text.length);
+				const outputWithoutInput = noModelLoading ? output : output.slice(text.length);
```
Pierrci, line 134: not sure how does it relate to `noModelLoading` here? ^^'

package.json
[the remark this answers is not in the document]
beurkinger: no, it won't work without it

js/src/lib/components/InferenceWidget/widgets/TextGenerationWidget/TextGenerationWidget.svelte
beurkinger: indeed, I'll look into it 👀

beurkinger: Ok I think that I did this because, OF COURSE, the output returned by Inference API in a couple of instances slightly different from the one returned by Inference Endpoints. So as `noModelLoading` is only used by Inference Endpoints, I used this to identify the API currently used, and then forgot to change this. I'll see what I can do - maybe add yet another param.

js/src/lib/components/InferenceWidget/shared/WidgetWrapper/WidgetWrapper.svelte
beurkinger: What's the logic behind hiding the widget while it's loading ? From what I saw if only create some nasty blinking effect when the widget suddenly appears.

beurkinger:
@Pierrci after considering https://github.com/huggingface/hub-docs/pull/916#discussion_r1356524501, instead of adding yet another prop, I choose to remove `appendRepoPath` and `noModelLoading`, and use a single `isInferenceEndpoints` prop instead. As these functionalities are only used in Inference Endpoints, I think it makes sense, and simplifies the codebase a bit.

js/src/lib/interfaces/Libraries.ts
beurkinger: Actually now that we're using hub-docs as a dependency, I don't think typescript is checking the files on our side. I don't mind removing these sort of changes.

I'll argue that as a library shared between different projects, and involving numerous contributors, maybe it would make sense to move toward stricter type checking options, as an easy way to help prevent unforeseen bugs and crashes.

beurkinger:
> Are you sure you want to go this way 😅
> Git submodule is hell for us (except for @julien-c)

We don't import it as a git submodules, but as a dependency in package.json ;)
see https://github.com/huggingface/hf-endpoints/blob/8a80bbf1347bb7b6f5b477c71c1873412f406e8f/front/package.json#L48C11-L48C11

The reason we choose to do it this way is that we use Vercel, and Vercel doesn't support Git Submodules (or we weren't able to get it to work in this instance, I can't remember). It has been working great for us, as we can easily pick a specific commit/branch by updating `package.json`, and we don't have the content of `hub-docs` polluting our codebase/typing.

js/src/lib/interfaces/Libraries.ts
julien-c: > we have a stricter tsconfig on the Endpoints side, with notably these params:
> 
> ```
>     "noUncheckedIndexedAccess": true,
>     "noImplicitAny": true,
> ```

i think we should align on the type-checking settings of moon-landing. So if you feel strongly about that, you should open a PR to change this config on moon-landing first

Pierrci: > Actually now that we're using hub-docs as a dependency, I don't think typescript is checking the files on our side. I don't mind removing these sort of changes from this PR.

Yes please then, `noUncheckedIndexedAccess` is a bit hardcore honestly (and wrong, in this case)

js/src/lib/components/InferenceWidget/shared/WidgetWrapper/WidgetWrapper.svelte
Pierrci: @mishig25

Pierrci:
> @Pierrci after considering [#916 (comment)](https://github.com/huggingface/hub-docs/pull/916#discussion_r1356524501), instead of adding yet another prop, I choose to remove `appendRepoPath` and `noModelLoading`, and use a single `isInferenceEndpoints` prop for everything. As these functionalities are only used in Inference Endpoints, I think it makes sense, and simplifies the codebase a bit.

I have very mixed feelings here; I'm not a fan of this (adding custom business logic with `isInferenceEndpoints`) but the alternative is not great either (adding _another_ ultra-specific prop like `sliceTextGenerationOutput` only used for one widget), so ok for me to go this way for now, and we can turn back interesting pieces of code to more lib-based logic later if needed.

js/src/lib/interfaces/Libraries.ts
beurkinger: @Pierrci it's not totally wrong because there are instances where the result of String.split() could be an empty array, but it's true that's not the case here :)

js/src/lib/components/InferenceWidget/shared/WidgetWrapper/WidgetWrapper.svelte
```diff
@@ -33,31 +34,37 @@
...
-	const inputSamples: WidgetInputSample[] = (model?.widgetData ?? [])
-		.sort((sample1, sample2) => (sample2.example_title ? 1 : 0) - (sample1.example_title ? 1 : 0))
-		.map((sample, idx) => ({
-			example_title: `Example ${++idx}`,
-			group: "Group 1",
-			...sample,
-		}));
+	let inputSamples: WidgetInputSample[] = [];
+	let inputGroups: { group: string; inputSamples: WidgetInputSample[] }[] = [];
+	$: {
```
beurkinger, line 39: Or else the widget won't update correctly when the props change (needed for hf-ndpoints)

js/src/lib/components/InferenceWidget/shared/types.ts
```diff
@@ -1,14 +1,16 @@
...
-	noTitle:            boolean;
-	shouldUpdateUrl:    boolean;
-	includeCredentials: boolean;
-	isLoggedIn?:        boolean;
+	accept?:               string;
+	apiToken?:             string;
+	apiUrl:                string;
+	callApiOnMount:        boolean;
+	isInferenceEndpoints?: boolean;
+	model:                 Pick<ModelData, 'id' | 'mask_token' | 'pipeline_tag' | 'widgetData'>;
```
beurkinger, line 9: Allow passing only the attributes that are actually needed - required for hf-endpoints. Doesn't change the widget behavior.

js/src/lib/components/InferenceWidget/widgets/FillMaskWidget/FillMaskWidget.svelte
```diff
@@ -142,7 +145,7 @@
 		<form>
 			{#if model.pipeline_tag === "fill-mask"}
 				<div class="mb-1.5 text-sm text-gray-500">
-					Mask token: <code>{model.mask_token}</code>
+					Mask token: <code>{model.mask_token ?? "unknown"}</code>
```
beurkinger, line 148: on the hf-endpoints side, we sometimes don't have the mast token available right away.

js/src/lib/inferenceSnippets/inputs.ts
```diff
@@ -44,7 +46,7 @@ const inputsTextGeneration = () => `"Can you please let us know more details abo
 
 const inputsText2TextGeneration = () => `"The answer to the universe is"`;
 
-const inputsFillMask = (model: ModelData) => `"The answer to the universe is ${model.mask_token}."`;
+const inputsFillMask = (model: ModelPartial) => `"The answer to the universe is ${model.mask_token ?? "[unknown_mask_token]"}."`;
```
beurkinger, line 49: on the hf-endpoints side, we sometimes don't have the mast token available right away.

PR #918, opened 2023-08-23
Support buttons in dropdown entry

krampstudio:
- Added a `type` property to change the clickable element type: either `<a>` or `<button>`
- If not set, we set it depending on the other properties (i.e. presence of `href`, etc.) for the best backward compat
- Also since the `DropdownMenu` has a max size, `truncate` entries by default

### DIFF
--- a/js/src/lib/components/DropdownEntry/DropdownEntry.svelte
+++ b/js/src/lib/components/DropdownEntry/DropdownEntry.svelte
@@ -24,7 +24,6 @@
 
 <li>
 	<svelte:element
-		this={element}
 		class="flex items-center w-full hover:bg-gray-50 dark:hover:bg-gray-800 cursor-pointer px-3 py-1.5 whitespace-nowrap
 		{classNames} {underline ? 'hover:underline' : ''}"
 		{href}
--- a/js/src/lib/components/DropdownEntry/DropdownEntry.svelte
+++ b/js/src/lib/components/DropdownEntry/DropdownEntry.svelte
@@ -24,6 +24,7 @@
 
 <li>
 	<svelte:element
+		this={element}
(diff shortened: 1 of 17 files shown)


mishig25 (approved):
lgtm!

js/src/lib/components/DropdownEntry/DropdownEntry.svelte
```diff
@@ -10,25 +10,40 @@
...
-		on:click={onClick}
-		rel={noFollow ? "nofollow" : undefined}
-		target={targetBlank ? "_blank" : undefined}
+		rel={element === "a" && noFollow ? "nofollow" : undefined}
+		target={element === "a" && targetBlank ? "_blank" : undefined}
+		on:click={(e) => {
+			if (type === "submit") {
+				e.stopPropagation();
+			}
+			onClick(e);
```
gary149, line 32: It would be better to forward a native event but probably out of scope of this PR (as we try to remove this callback pattern everywhere).

krampstudio: I was also tempted to move it to native events...

mishig25:
@krampstudio the js check is failing [here](https://github.com/huggingface/hub-docs/actions/runs/6024042329/job/16341962808). Would they be fixed as part of https://github.com/huggingface/hub-docs/pull/924 ? Thanks a lot !

krampstudio:
> @krampstudio the js check is failing [here](https://github.com/huggingface/hub-docs/actions/runs/6024042329/job/16341962808). Would they be fixed as part of #924 ? Thanks a lot !

yes, it fails because it doesn't recognize [svelte:element](https://github.com/huggingface/hub-docs/blob/main/js/src/lib/components/DropdownEntry/DropdownEntry.svelte#L21-L22) and it will be fixed by #924

PR #926, opened 2023-08-29
Fix sort param in models link

mkly:
Fix sort param in models link

*models* link results in 400 error "Invalid sort parameter". This is due to `sort=download` vs `sort=downloads`.

* Change to correct sort param in models link


julien-c (approved):

osanseviero (approved):
Thanks!

PR #933, opened 2023-08-31
Add `ctransformers` library

marella:
All the GGML/GGUF models from https://huggingface.co/TheBloke can be used with [ctransformers](https://github.com/marella/ctransformers).

### DIFF
--- a/docs/hub/_toctree.yml
+++ b/docs/hub/_toctree.yml
@@ -59,6 +59,8 @@
         title: AllenNLP
       - local: asteroid
         title: Asteroid
+      - local: ctransformers
+        title: CTransformers
       - local: espnet
         title: ESPnet
       - local: fastai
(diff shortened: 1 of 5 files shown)


osanseviero:
Hi there! At the moment, we have a strict 1 repo, 1 library structure on the Hugging Face Hub. For this to work, all of the repos you mention would have to have the `ctransformers` tag, otherwise they would not show. 

We see `GGUF`/`GGML` as file formats, the same as we see `safetensors`, `onnx`, etc. Because of this, we consider them framework-agnostic and don't enforce a library tag upon repos that have files with this format

julien-c:
yes, "library" in the Hub is really == modeling architecture library.

"File formats" however can coexist in the same repo, so maybe it would make sense to add some file formats for this

julien-c:
closing as stale, feel free to reopen

PR #942, opened 2023-09-08
Tweaks to connected apps docs

julien-c:
Follow-up to #902


julien-c:
And BTW we also need to add a doc page about the Connected Apps settings page

julien-c:
merging as i need to share it with a community member

PR #963, opened 2023-09-22
InferenceWidget stores file for svelte stores

mishig25:
Follow up to #961 

Good practice to have `stores.ts` file for svelte-stores rather than using `<script context="module"`


julien-c:
would be useful / good practice to ping someone else (familiar with this codebase) for review on those changes if possible

mishig25:
Will ask for reviewers on subseq PRs 👍

PR #968, opened 2023-09-26
Add udm to Language.ts

codemurt:
I recently noticed that the Udmurt language is not in the list of languages, although Komi, Mari and other related languages are already there. Therefore, I think it’s right to add the Udmurt language to this list.

Language code info: https://iso639-3.sil.org/code/udm
Used in dataset: https://huggingface.co/datasets/udmurtNLP/udmurt-bible-parallel-corpora


osanseviero (approved):
Thanks for this PR!

PR #970, opened 2023-09-27
Fix grammar, punctuation, and sentences in documentaion

nithishd-max:
It fixes the grammatical sentences and words and punctuation


osanseviero (approved):
Thanks!

PR #971, opened 2023-09-27
Tweaks to new spec for custom license stuff

julien-c:
Follow-up/tweak to #959


julien-c:
(and cc @apolinario  too)

PR #974, opened 2023-09-27
hmm 🤔

julien-c:
Screenshot 2023-09-27 at 20 06 07

PR #924, opened 2023-08-25
Align prettier config

krampstudio:
Same code style rules as in moon, and a CI that reflects it.

 Note:
 - if this PR is ok, we can `npm run format:all` + a few fixs and ignore the changes in the history

PR #975, opened 2023-09-27
Run linter

mishig25:
Should be merged into branch `why` #974

This PR:
1. Formats the files so that CI passes
2. Add eslint `alphabetize` rule https://github.com/huggingface/hub-docs/pull/975/commits/68c6c5ae044c6804c118013aebca19a7dc808fb4 so that you don't need to manually sort https://github.com/huggingface/hub-docs/pull/974/commits/12a3381b9a8b849cd8d8d93c7fd85e348d6524ec. However, we don't have this `alphabetize` in moon. And we are supposed to align with moon as written in https://github.com/huggingface/hub-docs/pull/924 @krampstudio wdyt?

### DIFF
--- a/js/.eslintrc.cjs
+++ b/js/.eslintrc.cjs
@@ -50,7 +50,6 @@ module.exports = {
 					"error",
 					{
 						"newlines-between": "always",
-						"alphabetize": { order: 'asc' },
 						"groups": ["type", "builtin", "external", "parent", "sibling", "index"],
 					},
 				],
--- a/js/.eslintrc.cjs
+++ b/js/.eslintrc.cjs
@@ -50,6 +50,7 @@ module.exports = {
 					"error",
 					{
 						"newlines-between": "always",
+						"alphabetize": { order: 'asc' },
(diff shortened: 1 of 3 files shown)


krampstudio:
>     2. Add eslint `alphabetize` rule [68c6c5a](https://github.com/huggingface/hub-docs/commit/68c6c5ae044c6804c118013aebca19a7dc808fb4) so that you don't need to manually sort [12a3381](https://github.com/huggingface/hub-docs/commit/12a3381b9a8b849cd8d8d93c7fd85e348d6524ec). However, we don't have this `alphabetize` in moon. And we are supposed to align with moon as written in [Align prettier config #924](https://github.com/huggingface/hub-docs/pull/924) @krampstudio wdyt?

eslint config can be specialized, so you can use the base rules from moon and add dedicated rules/plugins for hub_docs if it makes sense

julien-c:
don't think we really need `alphabetize` though

julien-c (approved):

PR #977, opened 2023-09-27
Fix curl code snippets

osanseviero:


Wauplin (approved):
Oupsi, thanks for fixing!

PR #979, opened 2023-09-28
Typings for all kinds of widget examples

julien-c:
still a draft, but pinging @mishig25 and @SBrandeis for a pre-review.

See #978 for wider context.

My main question right now is if i'm going to duplicate the validation (and types) in moon-landing in order to do `Joi` validation on the `cardData` – what do you think? (for now i'm thinking that i'll do handcoded validation because it's going to be too-complex to do in Joi)

### DIFF
--- a/js/src/lib/components/InferenceWidget/shared/WidgetExample.ts
+++ b/js/src/lib/components/InferenceWidget/shared/WidgetExample.ts
@@ -12,83 +12,78 @@ export interface WidgetExampleOutputText {
 export interface WidgetExampleOutputUrl {
 	url: string;
 }
+
+export type WidgetExampleOutput =
+	| WidgetExampleOutputLabels
+	| WidgetExampleOutputAnswerScore
+	| WidgetExampleOutputText
+	| WidgetExampleOutputUrl;
 //#endregion
 
-export interface WidgetExampleBase {
+export interface WidgetExampleBase<TOutput> {
 	example_title?: string;
 	group?:         string;
+	output?:        TOutput;
 }
 
-export interface WidgetExampleTextInputLabelsOutput extends WidgetExampleBase {
-	text:    string;
-	output?: WidgetExampleOutputLabels;
+export interface WidgetExampleTextInput<TOutput = WidgetExampleOutput> extends WidgetExampleBase<TOutput> {
+	text: string;
 }
 
-export interface WidgetExampleTextAndContextInputAnswerScoreOutput extends WidgetExampleBase {
-	text:    string;
+export interface WidgetExampleTextAndContextInput<TOutput = WidgetExampleOutput>
+	extends WidgetExampleTextInput<TOutput> {
 	context: string;
-	output?: WidgetExampleOutputAnswerScore;
 }
 
-export interface WidgetExampleTextInputTextOutput extends WidgetExampleBase {
-	text:    string;
-	output?: WidgetExampleOutputText;
+export interface WidgetExampleTextAndTableInput<TOutput = WidgetExampleOutput> extends WidgetExampleTextInput<TOutput> {
+	table: (string | number)[][];
 }
 
-export interface WidgetExampleTextInputUrlOutput extends WidgetExampleBase {
-	text:    string;
-	output?: WidgetExampleOutputUrl;
+export interface WidgetExampleAssetInput<TOutput = WidgetExampleOutput> extends WidgetExampleBase<TOutput> {
+	src: string;
+}
+export interface WidgetExampleAssetAndPromptInput<TOutput = WidgetExampleOutput>
+	extends WidgetExampleAssetInput<TOutput> {
+	prompt: string;
 }
 
-export interface WidgetExampleAssetInputLabelsOutput extends WidgetExampleBase {
-	src:     string;
-	output?: WidgetExampleOutputLabels;
-}
+export type WidgetExampleAssetAndTextInput<TOutput = WidgetExampleOutput> = WidgetExampleAssetInput<TOutput> &
+	WidgetExampleTextInput<TOutput>;
 
-export interface WidgetExampleAssetInputTextOutput extends WidgetExampleBase {
-	src:     string;
-	output?: WidgetExampleOutputText;
-}
+export type WidgetExampleAssetAndZeroShotInput<TOutput = WidgetExampleOutput> = WidgetExampleAssetInput<TOutput> &
+	WidgetExampleZeroShotTextInput<TOutput>;
 
-export interface WidgetExampleAssetInputUrlOutput extends WidgetExampleBase {
-	src:     string;
-	output?: WidgetExampleOutputUrl;
-}
-
-//#region more exotic stuff
-export interface WidgetExampleStructuredDataInputLabelsOutput extends WidgetExampleBase {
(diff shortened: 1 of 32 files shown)


js/src/lib/components/InferenceWidget/shared/WidgetExample.ts
```diff
@@ -0,0 +1,94 @@
...
+}
+export interface WidgetExampleOutputUrl {
+	url: string;
+}
+//#endregion
+
+export interface WidgetExampleBase {
+	example_title?: string;
+	group?:         string;
+}
```
mishig25, line 20:
```suggestion
export interface WidgetExampleBase<T> {
	example_title?: string;
	group?:         string;
	output?:        T;
}
```

since `output` attribute could be present in all the widget example types, maybe we can use generic here?

js/src/lib/components/InferenceWidget/shared/WidgetExample.ts
```diff
@@ -0,0 +1,94 @@
...
+
+export interface WidgetExampleTextInputLabelsOutput extends WidgetExampleBase {
```
mishig25, line 22: nit: naming wise we can remove the ending `Output`? Instead of `WidgetExampleTextInputLabelsOutput`, just `WidgetExampleTextInputLabels`

mishig25, line 22: nit: naming-wise we can just remove the ending `Output`. Instead of `WidgetExampleTextInputLabelsOutput`, just `WidgetExampleTextInputLabels`

js/src/lib/components/InferenceWidget/shared/WidgetExample.ts
```diff
@@ -0,0 +1,94 @@
+import type { TableData } from "./types";
+
+//#region outputs
+export type WidgetExampleOutputLabels = Array<{ label: string; score: number; }>;
```
mishig25, line 4:
```suggestion
export type WidgetExampleOutputLabels = Array<{ label: string; score: number; }>;
```

mishig25, line 4: btw, I'm not sure if there was ever agreement between using:
`Array<T>` vs `T[]`

mishig25:
> (for now i'm thinking that i'll do handcoded validation because it's going to be too-complex to do in Joi)

searching for a tool that automatically creates joi validation from ts interfaces. Unfortun, the [only one I found](https://github.com/pasientskyhosting/ts-joi-schema-generator) isn't maintained anymore

js/src/lib/components/InferenceWidget/shared/WidgetExample.ts
julien-c: cc @SBrandeis WDYT?

julien-c:
@mishig25 we do it in the other direction on moon-landing side: we write Joi validation and a tool generates types. It works well. 
Here as it's different repos, it will still require some copy/paste (because i don't want to add Joi as a dependency here), but i think that's fine.

js/src/lib/components/InferenceWidget/shared/WidgetExample.ts
julien-c: it's a bit less clear the distinction between input and output IMO

js/src/lib/components/InferenceWidget/shared/WidgetExample.ts
julien-c: not sure but here i copy/pasted this line from somewhere else

julien-c:
Ok, CI is green so i'm marking this as ready to review. 

ping @mishig25 and @SBrandeis for a final review as otherwise my brain might break 😩

js/src/lib/components/InferenceWidget/shared/WidgetExample.ts
julien-c: i tried but it didn't really simplify things.

That being said, i'm not great at types, so maybe you can try making it work better in a later PR?

js/src/lib/components/InferenceWidget/shared/WidgetWrapper/WidgetWrapper.svelte
```diff
@@ -1,6 +1,6 @@
-<script lang="ts">
+<script lang="ts" generics="T extends WidgetExample">
```
krampstudio, line 1: just wondering why a generic vs just `WidgetExample` here, since WidgetExample is already one of the defined type?

krampstudio (approved):
The new types looks good to me.
I've tested a couple of widgets, everything's good so far.

js/src/lib/components/InferenceWidget/shared/WidgetWrapper/WidgetWrapper.svelte
julien-c: i think the compiler complains if i do that (in the importing files)

js/src/lib/components/InferenceWidget/shared/WidgetExample.ts
mishig25: yeah that works! Therefore, resolving this issue for now

mishig25:
cool, let's merge #980 first

SBrandeis:
re-requesting review from @mishig25 @krampstudio now that #980 is merged

js/src/lib/components/InferenceWidget/shared/WidgetWrapper/WidgetWrapper.svelte
```diff
@@ -1,6 +1,6 @@
-<script lang="ts">
+<script lang="ts" generics="TWidgetExample extends WidgetExample">
```
mishig25, line 1: in the file [above](https://github.com/huggingface/hub-docs/pull/979/files#diff-db954a98b69d916d21056d16147e7c98bd981ec2174173d31d901f220991bd3e), we use `type TWidgetExample = $$Generic<WidgetExample>;`. Which style do we prefer?

mishig25 (approved):
lgtm except [this nit](https://github.com/huggingface/hub-docs/pull/979/files#r1341565539) !

js/src/lib/components/InferenceWidget/shared/WidgetWrapper/WidgetWrapper.svelte
krampstudio: I guess we should go for `ttype TWidgetExample = $$Generic<WidgetExample>;` because there's something wrong with the tooling (prettier or eslint remove the type import :/ )

julien-c: care to push this change directly to the PR?

krampstudio (approved):
lgtm :+1:

js/src/lib/components/InferenceWidget/shared/WidgetWrapper/WidgetWrapper.svelte
julien-c: e5cdad53b2867f80c849722dcf57742c8ad2981a

mishig25 (approved):
thanks for the push !

PR #982, opened 2023-09-29
Even more widgets tweaks

julien-c:
i reverted b473a9124edd80b3bc7c2ed2485c00fab1db925e bc it was a bit too quick n dirty, but @mishig25 let's keep it in mind for a potential future codebase evolution


julien-c:
Merging to unblock my moon-landing PR before the weekend 😇

PR #989, opened 2023-10-02
Hacktoberfest: new languages

stefan-it:
Hi,

I created some PRs that add language tag(s) to datasets for the Hacktoberfest :hugs: 

Both csv and readme files are updated.

Side note: I used PyCharm so the tables are perfectly padded and aligned as it is done in the original version!


davanstrien (approved):
Thanks again for all this work; this looks super nice :)

PR #996, opened 2023-10-04
[API docs] Replace `()` with `{}`

mishig25:
This PR is revert of commit https://github.com/huggingface/hub-docs/pull/962/commits/71898147b708d2e049f9a18e63715b908641a582.

Thanks to https://github.com/huggingface/doc-builder/pull/423, there will be no parsing errors when char `{` is inside a heading


Wauplin (approved):
Thanks a lot for fixing this @mishig25 ! Looks better this way

PR #1022, opened 2023-10-13
tiny inference wording suggestion

julien-c:
don't feel super strongly about it so up to you @osanseviero @Narsil @mishig25 i guess


osanseviero (approved):
LGTM!

PR #1029, opened 2023-10-15
Update about.md and sagemaker/getting-started.md

ShubhamJagtap2000:
There was a glitch in setting headlines for the first 2 lines in the `docs` section, I have edited it and now they are visible as they were expected. 

Also, in the https://github.com/huggingface/hub-docs/blob/main/docs/sagemaker/getting-started.md 
if you see the YouTube video frame is not available as expected as the <iframe> is no longer working with Markdowns.

So, I have added the YouTube video link for the Sagemaker Workshop with a frame as expected.

## Before changes screenshot

image

## After Changes screenshot

image

Please review.

Thank you!


julien-c:
Hmm, the video iframe does display for me on https://huggingface.co/docs/sagemaker/getting-started, not for you?

I agree with the header fixes though

tasks/src/image-classification/about.md
```diff
@@ -1,7 +1,7 @@
-## Use Cases
+## Use Cases
 Image classification models can be used when we are not interested in specific instances of objects with location information or their shape.
 
-### Keyword Classification
+### Keyword Classification
```
julien-c, line 4: thanks for this fix!

ShubhamJagtap2000:
> Hmm, the video iframe does display for me on https://huggingface.co/docs/sagemaker/getting-started, not for you?
> 
> I agree with the header fixes though

Oh, I was to tell about the in-GitHub documentation, @julien-c. 

In the docs, it was only the .md code being shown up, so I made it populate in GitHub MD file too.

Should we keep these changes in docs perspective or I should revert these changes and only keep header fixes? Will you be able to approve the PR?

Thank you.

julien-c:
yes, please keep the header fixes only. Thanks!

Issue #1028, opened 2023-10-13
Model Card: Clarification on linking models together


stefan-it:
Hi,

during reading the [Models FAQ](https://huggingface.co/docs/hub/models-faq) site, I came across the [following paragraph](https://huggingface.co/docs/hub/models-faq#what-if-i-have-a-different-checkpoint-of-the-model-trained-on-a-different-dataset):

![image](https://github.com/huggingface/hub-docs/assets/20651387/88360016-79c8-4352-98af-0fc2d7b9fb95)

It says:

```bash
You can link the models together by using a
[tag in your model card’s metadata](https://huggingface.co/docs/hub/modelcard)
or by linking to them in the model cards.
```

I can understand the second part (after the "or"). But I could not find any documentation for the linking of models together part via metadata tag. Unfortunately, the link is also broken. I could correct it to e.g. [`model-cards`](https://huggingface.co/docs/hub/model-cards) but how is linking of models possible?

julien-c:
I guess what that was trying to say was: you can pick a tag (any string keyword) related to your project and add it to `tags: ` in your metadata. Then by browsing `hf.co/models?filter=your_string` you'll find all of them

Another way now would be to use Collections as a way to group distinct related repos together.

julien-c:
(doc about Collections is at https://huggingface.co/docs/hub/collections)

stefan-it:
Thanks for these explanations @julien-c .

I opened a PR that extends the FAQ a bit (incl. adding a reference to the new awesome Collections feature), see: #1031

PR #1031, opened 2023-10-17
docs: add more clarifications about how to link models together

stefan-it:
Hey,

related to #1028 I added some more explanations about linking models together:

* by specifying a tag in `tags` array in the model card's metadata, and
* by using the new Collection feature

:hugs:


julien-c (approved):
thanks!

osanseviero (approved):
Thank you!

PR #1045, opened 2023-10-23
added language to Dataset Card for M3IT#986

smty2018:
added language to Dataset Card for M3IT
#986


Wauplin (approved):
Thanks you for your contribution @smty2018! Will merge now :)

PR #1046, opened 2023-10-23
Update datasets_without_language.md #986

smty2018:
Added language to dataset card huggan/wikiart 
 #986


Wauplin (approved):
Thanks you for this new contribution @smty2018! :hugs:

PR #1057, opened 2023-10-27

Ashish-Soni08:
model_id -> huggyllama/llama-30b


Wauplin (approved):
Looks good!

PR #1059, opened 2023-10-27

Ashish-Soni08:
model_id -> decapoda-research/llama-13b-hf


Wauplin (commented):
Weird, looks like https://huggingface.co/decapoda-research/llama-13b-hf doesn't exist anymore (most likely to be private :confused:)

PR #1060, opened 2023-10-27

Ashish-Soni08:
model_id -> TheBloke/Llama-2-70B-fp16


Wauplin (changes requested):
Llama2 models are a special case. You can directly update to `license: llama2` in the metadata. No need for `license_name`/`license_link` then :)

See in https://github.com/huggingface/hub-docs/blob/main/hacktoberfest_challenges/model_license_other.md

image

Ashish-Soni08:
Sorry, I made the change :)

Wauplin (approved):
Awesome! Thanks for making the change :pray: And PR is even merged, that's perfect! :fire:

PR #1061, opened 2023-10-27

Ashish-Soni08:
model_id -> sambanovasystems/BLOOMChat-176B-v1


Wauplin (commented):
For simplicity, I would rename it to `bloomchat-v1.0` instead of `bloomchat-176b-license-v1.0`. What do you think?

Wauplin (approved):
Nice! Thanks for making the change :)

Issue #1011, opened 2023-10-10
Show static widget output on load


osanseviero:
Sometimes the widget output is shown and sometimes it's not shown on page load. @mishig25 hinted it could be a `onload` race condition

PR #1023, opened 2023-10-13
[Widgets] Refactor examples running

mishig25:
## Refactor codes related to running widget examples

Overall, this refactor cleans some code & adds more usage of typing.

1. Replaces `getDemoInputs` with `getWidgetExample` function as proposed in #1021
2. Removes `previewInputSample` functions entirely by re-using `applyInputSample` function with new `isPreview` opts. (See more [here](https://github.com/huggingface/hub-docs/pull/1017#discussion_r1358032660))

change signature of 
```ts
function applyInputSample(sample: WidgetExample)
```
to
```ts
function applyInputSample(sample: WidgetExample, opts: {isPreview?: boolean, inferenceOpts?: InferenceRunOpts; })
```

so that [applyInputSample](https://github.com/huggingface/hub-docs/blob/b2ba06d025a6de85049e6470f1533f6f59b4e9e8/js/src/lib/components/InferenceWidget/widgets/AudioClassificationWidget/AudioClassificationWidget.svelte#L125) fully replaces [previewInputSample](https://github.com/huggingface/hub-docs/blob/b2ba06d025a6de85049e6470f1533f6f59b4e9e8/js/src/lib/components/InferenceWidget/widgets/AudioClassificationWidget/AudioClassificationWidget.svelte#L133) when used with `opts isPreview=true` and [applyInputSample with getWidgetExample](https://github.com/huggingface/hub-docs/blob/b2ba06d025a6de85049e6470f1533f6f59b4e9e8/js/src/lib/components/InferenceWidget/widgets/AudioClassificationWidget/AudioClassificationWidget.svelte#L152-L155) replaces [getDemoInputs and running example code in onMount](https://github.com/huggingface/hub-docs/blob/b2ba06d025a6de85049e6470f1533f6f59b4e9e8/js/src/lib/components/InferenceWidget/widgets/AudioClassificationWidget/AudioClassificationWidget.svelte#L150-L156). (in this example, I've used `AudioClassificationWidget.svelte` but the changes apply to all widgets).

### Order of merge operations
1. Merge #1023 (this PR)
2. Then merge #1017
3. Then merge #1019

PR #1019, opened 2023-10-11
[Widgets] Handle display logic

mishig25:
Closes https://github.com/huggingface/hub-docs/issues/1010

### This is how it looks like, when an inference api is disabled for a model.

image

### This is how it looks like, when an inference api is disabled for a model BUT there is an example with output (#979).
Detailed description:
1. There will be message that says `However, you can still play with [examples that have an output](hf.co/docs/hub/models-widgets#example-outputs)`
2. All the input elements are disabled (you can't enter text, the compute button is greyed out)
3. As a user, you only see examples that have outputs. If in the readme, there was an example without output, it is being filtered out.
4. In the current implementation, when you click an example with an output, it tries to call an inference-api. This should be fixed with #1017

image

### How to test

```
cd js
npm ci
npm run dev -- --open
```
and the command above should open a webpage

### Order of merge operations
1. Merge #1023 
2. Then merge #1063 
3. Then merge #1019 (this PR)

PR #1017, opened 2023-10-11
[Widgets] Don't call api inference if output exists

mishig25:
Closes https://github.com/huggingface/hub-docs/issues/1011
Follow up to https://github.com/huggingface/hub-docs/pull/979/

#979 correctly implemented behaviour for the [previewInputSample function here](https://github.com/huggingface/hub-docs/blob/7cfc6cd525b1f726e7394a95947f96b42d1082b3/js/src/lib/components/InferenceWidget/widgets/AudioClassificationWidget/AudioClassificationWidget.svelte#L149-L151). Which is: don't call api-inference if `sample.output` exists

The same should be the case for [`applyInputSample` and `onMount` example](https://github.com/huggingface/hub-docs/pull/1017/files) (which is this PR). Don't call api-inference if `sample.output` exists.  https://github.com/huggingface/hub-docs/issues/1011 happened because api-inference was called and there was no cached example although `sample.output` was in the repo

Wdyt of the proposed solution? I've only implemented for [AudioClassificationWidget.svelte](https://github.com/huggingface/hub-docs/pull/1017/files#diff-a33c2021426ab67a2b34852d3250eb722214cd4e21f6de67d70ac8cdc9ed1b2a). If others agree with the design, will implement for other widgets as well

** for the clarification, `previewInputSample` function gets called when you hover over different input samples. `applyInputSample` function gets called when you end up clicking/selecting inout sample

### Order of merge operations
1. Merge #1023 
2. Then merge #1017 (this PR)
3. Then merge #1019

PR #1063, opened 2023-10-30
[Widgets] Don't call api inference if output exists

mishig25:
Closes https://github.com/huggingface/hub-docs/issues/1011
This PR is continuation of #1017
Follow up to https://github.com/huggingface/hub-docs/pull/979/

#979 correctly implemented behaviour for the [previewInputSample function here](https://github.com/huggingface/hub-docs/blob/7cfc6cd525b1f726e7394a95947f96b42d1082b3/js/src/lib/components/InferenceWidget/widgets/AudioClassificationWidget/AudioClassificationWidget.svelte#L149-L151). Which is: don't call api-inference if `sample.output` exists

The same should be the case for `applyInputSample` (which is this PR). Don't call api-inference if `sample.output` exists.  https://github.com/huggingface/hub-docs/issues/1011 happened because api-inference was called and there was no cached example although `sample.output` was in the repo

### Order of merge operations
1. Merge #1023 
2. Then merge #1063 (this PR)
3. Then merge #1019

### DIFF
--- a/js/src/lib/components/InferenceWidget/widgets/AudioToAudioWidget/AudioToAudioWidget.svelte
+++ b/js/src/lib/components/InferenceWidget/widgets/AudioToAudioWidget/AudioToAudioWidget.svelte
@@ -62,7 +62,11 @@
 		}
 	}
 
-	async function getOutput({ withModelLoading = false, isOnLoadCall = false }: InferenceRunOpts = {}) {
+	async function getOutput({
+		withModelLoading = false,
+		isOnLoadCall = false,
+		exampleOutput = undefined,
+	}: InferenceRunOpts = {}) {
 		if (!file && !selectedSampleUrl) {
 			error = "You must select or record an audio file";
 			return;
@@ -134,7 +138,8 @@
 		}
 		file = null;
 		selectedSampleUrl = sample.src;
-		getOutput(opts.inferenceOpts);
+		const exampleOutput = sample.output;
+		getOutput({ ...opts.inferenceOpts, exampleOutput });
 	}
 </script>
 
--- a/js/src/lib/components/InferenceWidget/widgets/AudioToAudioWidget/AudioToAudioWidget.svelte
+++ b/js/src/lib/components/InferenceWidget/widgets/AudioToAudioWidget/AudioToAudioWidget.svelte
@@ -1,5 +1,5 @@
 <script lang="ts">
-	import type { WidgetProps, ExampleRunOpts, InferenceRunFlags } from "../../shared/types";
+	import type { WidgetProps, ExampleRunOpts, InferenceRunOpts } from "../../shared/types";
 	import type { WidgetExampleAssetInput } from "../../shared/WidgetExample";
 
 	import WidgetAudioTrack from "../../shared/WidgetAudioTrack/WidgetAudioTrack.svelte";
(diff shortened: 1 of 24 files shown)


js/src/lib/components/InferenceWidget/widgets/AudioToAudioWidget/AudioToAudioWidget.svelte
```diff
@@ -62,7 +62,11 @@
 		}
 	}
 
-	async function getOutput({ withModelLoading = false, isOnLoadCall = false }: InferenceRunFlags = {}) {
+	async function getOutput({
+		withModelLoading = false,
+		isOnLoadCall = false,
+		exampleOutput = undefined,
```
SBrandeis, line 68: This has no effect in this widget

SBrandeis (commented):
The logic is not implemented for most widgets. Is that intended?

js/src/lib/components/InferenceWidget/widgets/AudioToAudioWidget/AudioToAudioWidget.svelte
mishig25: yes, because widgets such as AudioToAudio or ImageSegmentation are not expected to have `example.output` in their model cards

for example here https://github.com/huggingface/hub-docs/blob/e5cdf0af4ea36411f489c290e6af144e43c7723c/js/src/lib/components/InferenceWidget/widgets/AudioToAudioWidget/AudioToAudioWidget.svelte#L131, if they were expected to have `example.output`, the typing would have been `sample: WidgetExampleAssetInput<WIDGET OUTPUT TYPE>`

docs here for the list of supported output types: https://huggingface.co/docs/hub/models-widgets#example-outputs from #978

@julien-c could you confirm this statement
> yes, because widgets such as AudioToAudio or ImageSegmentation are not expected to have `example.output` in their model cards

julien-c: i'm not sure i understand, i'm a bit lost...

julien-c: from the docs (emphasis mine):
> **each** widget example can also optionally describe the corresponding model output, directly in the output property

AudioToAudio or ImageSegmentation should support a file output, i.e. a `output.url`, same as text-to-image:

image

mishig25: As we discussed internally, lets merge this PR without the proposed changes above (`remaining widgets validate and use output`). And iterate in subseq PRs

SBrandeis (approved):

js/src/lib/components/InferenceWidget/widgets/AudioToAudioWidget/AudioToAudioWidget.svelte
julien-c: > Should I open a PR (extend the current PR) to make all the remaining widgets validate and use `output` ?

Yes

julien-c: that was the goal to support `output` for all widgets

PR #1064, opened 2023-10-30
`structuredData` -> `structured_data`

mishig25:


julien-c:
hmm ok but you need to open PRs to change the models on the Hub then (there are not that many of those, so should be doable)

e.g. https://huggingface.co/julien-c/wine-quality/blob/main/README.md?code=true#L9

mishig25:
>  you need to open PRs to change the models on the Hub 

will do so

julien-c:
Note that you also have occurences of `structuredData` in the moon-landing codebase.... can you please open a PR there?

julien-c:
there are even some remaining occurences inside hub-docs!

mishig25:
> Note that you also have occurences of structuredData in the moon-landing codebase

opened a companion PR

> there are even some remaining occurences inside hub-docs!

I might be missing something. I don't see any in this search [here](https://github.com/search?q=repo%3Ahuggingface%2Fhub-docs%20structuredData&type=code).
There is this line (note the typo `strucured` vs `structured` PR #1069). In this case, this variable will not cause any problem because it has no relation to the query param `strucured_data`. So I don't think there is any remaining occurences inside hub-docs. Correct me if I'm wrong

https://github.com/huggingface/hub-docs/blob/cd15b19ac748517845b00d1aa49dd2e5370835d2/js/src/lib/components/InferenceWidget/widgets/TabularDataWidget/TabularDataWidget.svelte#L45

julien-c:
There's one in index.svelte

julien-c:
oh nvm looks like it was changed

PR #1067, opened 2023-10-30
[Widgets refactor] typing use `enum`s

mishig25:
Nit: use enums when makes sense here


SBrandeis (commented):
What's the benefit of typing it this way?

mishig25:
No benefit. I thought it would be stylistically consisted with https://github.com/huggingface/hub-docs/blob/5e3dcd184361d9552ea4c6f920fb49c23a3f690d/js/src/lib/interfaces/InferenceDisplayability.ts#L1-L14

SBrandeis (approved):

PR #905, opened 2023-07-28
Fix example latex

Kakulukian:
Fixing the misleading example for LaTex, used a workaround with html entities to avoid the parsing

PR #1074, opened 2023-11-02
try to fix current issue with latex inline doc

julien-c:
current way it looks:

image

Follow-up to #905?


Kakulukian (approved):

mishig25:
fixed by https://github.com/huggingface/doc-builder/pull/438

https://huggingface.co/docs/hub/model-cards#can-i-write-latex-in-my-model-card

image

Issue #193, opened 2022-06-09
Issue about languages tags


lbourdois:
Hi,

A PR to discuss an issue raised in an exchange following a Hub PR : [https://huggingface.co/datasets/AmazonScience/massive](https://huggingface.co/datasets/AmazonScience/massive)/discussions/1  (cc @julien-c).

The author of the dataset (the logic described below also applies to models) has filled language tags in the form **"xx-XX"** (e.g. fr-FR). However, the filters on the Datasets page (https://huggingface.co/datasets) only allow filtering by language tags of the form "**xx**" (e.g. "fr"). This meant that its dataset was not findable via the filters before the Hub PR added language tags of the form "xx".

This is something that happens quite frequently. A far from exhaustive example:
• 59 "en-US" datasets not counted in the "en" ([https://huggingface.co/datasets?languages=languages:en-US](https://huggingface.co/datasets?languages=languages:en-US))
• 24 "zh-CN" datasets not counted in the "zh" ([https://huggingface.co/datasets?languages=languages:zh-CN](https://huggingface.co/datasets?languages=languages:zh-CN))
• 20 "sv-SE" datasets not counted in the "sv" ([https://huggingface.co/datasets?languages=languages:sv-SE](https://huggingface.co/datasets?languages=languages:sv-SE))
• 15 "rm-sursilv" dataset not counted in the "rm" ([https://huggingface.co/datasets?languages=languages:rm-sursilv](https://huggingface.co/datasets?languages=languages:rm-sursilv))
• 15 "ga-IE" dataset not counted in the "ga" ([https://huggingface.co/datasets?languages=languages:ga-IE](https://huggingface.co/datasets?languages=languages:ga-IE))
• 13 "fy-NL" dataset not counted in the "fy" ([https://huggingface.co/datasets?languages=languages:fy-NL](https://huggingface.co/datasets?languages=languages:fy-NL))
• 10 "pa-IN" dataset not counted in the "pa" ([https://huggingface.co/datasets?languages=languages:pa-IN](https://huggingface.co/datasets?languages=languages:pa-IN))
• 5 "de-DE" dataset not counted in the "de" ([https://huggingface.co/datasets?languages=languages:de-DE](https://huggingface.co/datasets?languages=languages:de-DE))
Out of these 8 languages, 161 datasets are not findable via the language filter (actually a little less because there are overlapping multilingual datasets).

[...]

PR #1082, opened 2023-11-05
Add all ISO-639-3 languages' names to Language.ts

lbourdois:
Hi!

I was wandering around the Hub and came across https://huggingface.co/jbochi/madlad400-3b-mt, which offers over 400 languages.
Looking at them, it turns out that many have only the ISO-639-3 tag displayed but not their name.

image

I therefore propose this PR adding all the ISO-639-3 tag names (the data comes from the official SIL International website: https://iso639-3.sil.org/sites/iso639-3/files/downloads/iso-639-3.tab).
Languages already present in the `Language.ts` file are not affected by this PR.
I add the `"code"` and the `"name"` each time, but not the `"nativeName"` which is not present in the SIL file.
I suppose people will open PRs to add the native name of a language if they express the need. If Wikipedia is well structured, I might be able to scrape it to retrieve these native names from the language pages, but that would be for another PR. 

Let me know if there's anything to change.

### DIFF
--- a/js/src/lib/interfaces/Language.ts
+++ b/js/src/lib/interfaces/Language.ts
@@ -8,7 +8,7 @@ export function wikiLink(l: Language): string {
 	return `https://en.wikipedia.org/wiki/ISO_639:${l.code}`;
 }
 
-export const LANGUAGES: Record<string, Language> = {
+export const LANGUAGES_ISO_639_1: Record<string, Language> = {
 	aa: {
 		code:       "aa",
 		name:       "Afar",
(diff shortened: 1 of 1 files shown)


julien-c:
hi @lbourdois, `lzh` and `udm` were already listed so they're both listed twice now

julien-c:
oh actually can you generate a separate dict (same format) that includes *all* the ISO-639-3 tags? I'll rework how we merge the two dicts (in a separate PR)

lbourdois:
@julien-c I deleted the `lzh` and `udm` tags that were duplicated.

Concerning the generation of a dictionary including all ISO-639-3 tags, two questions:
- Do you want me to generate absolutely all languages, i.e. languages currently based on the ISO-639-1 tag in the `Language.ts` file should also be in ISO-639-3 (for example, French will be tagged `fra` instead of `fr`)?
And then you'll do your own code to either revert to ISO-639-1 for those languages, or something that allows those languages to handle both 639-1 and 639-3 tags.
- Should I put this generated dictionary in `Language.ts` or elsewhere?

julien-c:
just to understand better, in your current PR, did you manually remove the tags like `fra` which cover a language that's already in ISO-639-1?

lbourdois:
Absolutely

lbourdois:
This PR is for ISO-639 codes, but I've just thought it would be quite easy to add the glottocodes used by linguists, and thus encourage them to put their data on the Hub.
The dictionary would then look something like:

```
Abinomn: {
		code: "bsa",
		glottocode: "abin1243",
		near_code: "None",
		name: "Abinomn"
	},
Gallo: {
		code: "None",
		glottocode: "gall1275",
		near_code: "fra",
		name: "Gallo"
	},
```
In this case, we use the name of the language in English as the key. 
In value we have `code`, which corresponds to what HF currently uses (= the ISO codes), `glottocode` corresponds to the equivalent of the ISO code but for the glottocode convention, `near_code` corresponds to the ISO code closest to the language when there is no corresponding ISO code in the language (we have around 7,000 ISO code versus over 20,000 glottocode, so many more languages/dialects taken into account, it also tells us which languages are dialects of a main language); and finally `name` the name of the language.

If you're interested in this approach, I've put the complete dictionary here: https://github.com/lbourdois/iso-639-3/blob/main/Language_with_glottocode

julien-c:
BTW linking this PR back to #193 and https://github.com/huggingface/datasets/issues/4881 among others

julien-c:
@lbourdois honestly i feel like `glottocode` are a little bit too complex (for our use-case) and i think it's already an improvement if we have the full ISO-639-3 DB in this file (which i'll move to a https://github.com/huggingface/huggingface.js subpackage if we move forward with that PR)

Theoritically this will already close https://github.com/huggingface/datasets/issues/4881

Are you ok if i move forward with this plan? (merge your PR, then move the file to the huggingface.js repo, and use it in the Hub)

lbourdois:
@julien-c I suppose it's better than nothing ; `glottocode` is indeed complicated and not very intuitive.
The solution would rather be to complain to ISO that they're missing a bunch of languages from their list and therefore need to add new tags. I'd probably do that 👀

So no problem moving forward.
I just point out that I''ve got some code running right now that retrieves the native names of ISO-639-3 languages from wikipedia. 
So if I'm not mistaken and it doesn't crash overnight, I'll be able to add that to this PR tomorrow.
According to my first rough estimates, this should make it possible to add the native name to 15 to 20% of ISO languages (for the others, the information is simply not available on wikipedia).

julien-c:
honestly i don't think we need native names (we almost don't use them in the Hub)

julien-c:
I'll tweak this PR today, merge it, then move it to https://github.com/huggingface/huggingface.js as discussed.

I think we'll be done for language integration in the Hub 🎉 Thanks for your help @lbourdois

julien-c (approved):

lbourdois:
@julien-c 
If it can ever be useful, I've added the clean dataset of 1791 native names here:  https://github.com/lbourdois/iso-639-3/blob/main/native_names.csv

PR #1084, opened 2023-11-06
change metadata key from `finetuned_from` to `base_model`

julien-c:
see internal discussion there https://huggingface.slack.com/archives/C01BWJU0YKW/p1687766681122359

related to (internal) https://github.com/huggingface/moon-landing/pull/7989

### DIFF
--- a/docs/hub/model-card-annotated.md
+++ b/docs/hub/model-card-annotated.md
@@ -81,7 +81,7 @@ _Name and link to the license being used._
 
 
 
-* **Finetuned From Model [optional]:** `finetuned_from`
+* **Finetuned From Model [optional]:** `base_model`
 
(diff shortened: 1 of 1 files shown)


meg-huggingface:
Fine with me, although what if the model is fine-tuned from a model that's
been fine-tuned (is not the base model)?

Agree that `base_model` should be a metadata tag, though.

julien-c:
@meg-huggingface good point, in our spec base_model is the direct "parent" so the keyword is a bit confusing vs. the transitive "topmost" base model 

The good news is that we can compute the "topmost" grand-grand-[...]-parent quite easily from the HF API

osanseviero (approved):

PR #1087, opened 2023-11-07
Fix link

meg-huggingface:
Kept getting requests for a WIP google doc, and could not figure out where they were coming from -- realized that the doc was still linked to here, and so needed to be fixed.


EziOzoani (approved):
Aww no 
Thank you for catching that

PR #1088, opened 2023-11-07
Proposal: refactor widgets

mishig25:


js/src/lib/components/InferenceWidget/shared/types.ts
```diff
@@ -66,3 +66,20 @@ export interface ImageSegment {
...
+
+export interface WidgetInput {
+	// example functions
+	validateExample:   <TOutput>(sample: WidgetExample<TOutput>) => sample is WidgetExample<TOutput>;
+	applyExample:      (sample: any, opts: ExampleRunOpts) => void;
+	// ingerence api functions
+	getInferenceInput: () => Promise<any>;
+}
+
+export interface WidgetOutput {
```
mishig25, line 78: here, i tried to explicitly differentiate between `validateExample` and `showInferenceOutput`. On most cases, these two functions should be identical. However, there are instances where it can differ, [isValidOutput](https://github.com/huggingface/hub-docs/blob/6cc444ef92c35788bdf24791bb2ee4f30d35783e/js/src/lib/components/InferenceWidget/widgets/AudioToAudioWidget/AudioToAudioWidget.svelte#L115) is different from [isValidOutputUrl](https://github.com/huggingface/hub-docs/blob/b2ba06d025a6de85049e6470f1533f6f59b4e9e8/js/src/lib/components/InferenceWidget/shared/outputValidation.ts#L20)

SBrandeis:
Reviewing this today!

mishig25:
Thanks. It is a proposal. So I only implemented fkr AudioClassification widget. And if/when we agree on decisions, will implement for all widgets

mishig25:
Thanks. It is a proposal. So I only implemented fkr AudioClassification widget. And if/when we agree on decisions, will implement for all widgets

js/src/lib/components/InferenceWidget/shared/WidgetWrapperV2/WidgetWrapperV2.svelte
```diff
@@ -0,0 +1,218 @@
...
+	export let apiToken: WidgetProps["apiToken"];
+	export let isLoading = false;
+	export let model: WidgetProps["model"];
+	export let includeCredentials: WidgetProps["includeCredentials"];
+	export let noTitle = false;
+	export let exampleQueryParams: WidgetExampleAttribute[] = [];
+
+	export let widgetInput: WidgetInput;
+	export let widgetOutput: WidgetOutput;
+
```
SBrandeis, line 31: It's very unclear to me who/what sets the `WidgetInput`/`WidgetOutput` info here

js/src/lib/components/InferenceWidget/widgets/AudioClassificationWidgetV2/AudioClassificationWidgetV2.svelte
```diff
@@ -0,0 +1,39 @@
...
+	{widgetOutput}
+	bind:getOutput
+>
+	<svelte:fragment slot="input" let:isDisabled>
+		<WidgetInputAudio bind:widgetInput {isLoading} {isDisabled} on:click={() => getOutput()} />
+	</svelte:fragment>
+	<svelte:fragment slot="output">
+		<WidgetOutputClassification bind:widgetOutput />
+	</svelte:fragment>
+</WidgetWrapper>
```
SBrandeis, line 39: This is nice :+1:

I think it makes sense to fatcorize into inputs and outputs groups

js/src/lib/components/InferenceWidget/shared/WidgetWrapperV2/WidgetWrapperV2.svelte
SBrandeis: From what I understand:
- it is set by the `Input` or `Output` svelte components (which is sane - hold the validation logic close to the display logic)
- It goes all the way up the component hierarchy to `WidgetWrapper`, which uses it to filter and validate the  examples from model card

I think that:
1. the bottom-up data flow is not idiomatic and confusing.
2. Validation logic gets split between `WidgetWrapper` and the input/output components, which kinda breaks encapsulation

SBrandeis (commented):
While I like the inpiut / output factorization, I think we're trying to do too much stuff in WidgetWrapper which hurts readability and encapsulation.

mishig25:
```html
// WidgetWrapper
<script>
    export let WidgetInputComponent;
    let widgetInput;

    export let WidgetOutputComponent;
    let widgetOutput:

    let error = ""

    function validateExample(){
        widgetInput.validateExample() && widgetOutput.validateExample()
    }

    function callApi(){
        reqBody = widgetInput.getReqBody();
        output = callInferenceApi(reqBody)
        widgetOutput.renderOutput(output)
    }
</script>

<WidgetHeader/>

<WidgetInputComponent bind:widgetInput/>
<WidgetOutputComponent bind:widgetOutput/>

<WidgetFooter/>
```

mishig25:
> @huggingface/inference, we probably won't need a validateOutput function anymore?

oh indeed. It would make sense to use @huggingface/inference then rather than trying to reinvent validation here. I will look into it

PR #1089, opened 2023-11-07
[Hotfix] Enterprise docs

mishig25:
At the moment, doc-builder uses markdown headings for creating right-side nav (see attached img for example)

image

I will work on supporting html headings for right-side nav. Until then, this is a hotfix


julien-c (approved):
what's super weird is that it worked in CI, at least when partial-loading (didn't work when doing a full page load)

PR #1091, opened 2023-11-08
Add notice for "createdAt" key in API

Kakulukian:
For https://github.com/huggingface/moon-landing/pull/7991 (private)

### DIFF
--- a/docs/hub/api.md
+++ b/docs/hub/api.md
@@ -152,11 +152,6 @@ Get all information for a specific model.
 
 This is equivalent to `huggingface_hub.space_info(repo_id, revision)`.
 
-### GET /api/metrics
-
-Get information from all metrics in the Hub.
-
-This is equivalent to `huggingface_hub.list_metrics()`.
 
 ## Repo API
 
--- a/docs/hub/api.md
+++ b/docs/hub/api.md
@@ -13,7 +13,6 @@ Try it out now on our [Playground](https://huggingface.co/spaces/enzostvs/hub-ap
 <img class="w-full object-contain" src="https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/hub/hub-api-playground.png"/>
 </div>
 
-
 ## Repo listing API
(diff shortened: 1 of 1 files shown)


docs/hub/api.md
```diff
@@ -142,11 +150,20 @@ params = {
...
 
 This is equivalent to `huggingface_hub.space_info(repo_id, revision)`.
 
+The key `createdAt` in the response begins on `2022-03-02T23:29:04.000Z` as a result of a special migration.
+
+### GET /api/metrics
+
+Get information from all metrics in the Hub.
+
+This is equivalent to `huggingface_hub.list_metrics()`.
```
julien-c, line 166: stale addition

julien-c: (see PR diff)

docs/hub/api.md
```diff
@@ -46,12 +46,16 @@ params = {
 
 This is equivalent to `huggingface_hub.list_models()`.
 
+The key `createdAt` in the response begins on `2022-03-02T23:29:04.000Z` as a result of a special migration.
```
julien-c, line 49: IMO this should be mentioned just once, under a `# Repo creation times` heading, with a wording like:
> For all repos-returning API calls, the `createdAt` attribute is the time when the repo was created.
> Note that there is a special value of `2022-03-02T23:29:04.000Z` for all repos that were created before we started storing the creation date.

cc'ing @Wauplin too

julien-c (changes requested):

docs/hub/api.md
Kakulukian: Yeah sure

julien-c (approved):
lgtm but wdyt @Wauplin?

Wauplin (approved):
Looks good! Will integrate it to `huggingface_hub` as well :fire:

julien-c:
looks fine in prod IMO, thanks:

image

PR #1098, opened 2023-11-15
Update ZenML Spaces docs

strickvl:
ZenML recently changed some things about how the dashboard issues / handles cookies and so currently the dashboard doesn't work from within the Huggingface website wrapper (only via the direct URL).

I've amended the documentation to reflect this and to guide people in how to get it working.

### DIFF
--- a/docs/hub/spaces-sdks-docker-zenml.md
+++ b/docs/hub/spaces-sdks-docker-zenml.md
@@ -61,28 +61,26 @@ After creating your Space, you'll notice a 'Building' status along with logs
 displayed on the screen. When this switches to 'Running', your Space is ready for use. If the
 ZenML login UI isn't visible, try refreshing the page.
 
-Use our default login to access the dashboard (username: 'default', password:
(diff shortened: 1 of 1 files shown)


davanstrien (approved):
LGTM but just pinging @osanseviero to see if they know of any other workarounds for the cookies issue.

strickvl:
Yeah the problem is that we assign a cookie when you log in with the direct URL (which uses that domain for the cookie etc), but then that cookie isn't valid for the `huggingface.co` domain. We considered relaxing our cookie security / protocols etc, but that also doesn't seem satisfactory.

davanstrien:
@strickvl, that makes sense; we're just checking to see if there are any other workarounds from our end. @osanseviero I suggest we merge this for now and update if we find another approach?

osanseviero (approved):
Sounds good, ideally this should work in HF page, I shared this issue internally. This is specially important here as it's a Docker Template and many users will land in a confusing experience otherwise

PR #1100, opened 2023-11-15
rm Dropdown

julien-c:


mishig25 (approved):

julien-c:
@beurkinger i think you will want to copy/paste `Dropdown` into the inference-endpoints codebase as it will decouple the two projects a bit more

PR #1111, opened 2023-11-21
Added mask-generation and zero-shot-object-detection

merveenoyan:
This PR adds mask-generation and zero-shot-object-detection as tasks on Hub.


merveenoyan:
Once merged we can let people contribute the task page, so I've put placeholder over there.

osanseviero (changes requested):
All of this is being moved to https://github.com/huggingface/huggingface.js as discussed in Slack, so I would wait for it to be there and do the changes there (cc @coyotte508 ). E.g. the tasks codebase is already over there

osanseviero:
Actually everything is there already, so we can apply changes there (e.g. https://github.com/huggingface/huggingface.js/blob/ccfb044414a235270cb82d8b586f81785ad04eba/packages/tasks/src/pipelines.ts#L2)

PR #1135, opened 2023-11-25
Update docs to use references in huggingface.js

mishig25:
Follow up to https://github.com/huggingface/huggingface.js/pull/296 & https://github.com/huggingface/huggingface.js/pull/297

PR #1136, opened 2023-11-26
Standardize references to file

mishig25:
Nit: I've noticed that references to files could be standardized bit more while working on #1135


merveenoyan:
On tasks' end I'm in favor of keeping existing ones for the sake of readability, unless we have any other objection.

EziOzoani (commented):
I like the approach; it reads a lot cleaner 
Thanks @mishig25 🤗

osanseviero (approved):
LGTM from a quick review

docs/hub/models-adding-libraries.md
```diff
@@ -184,7 +184,7 @@ All third-party libraries are Dockerized, so you can install the dependencies yo
 
 ### Register your libraries supported tasks on the hub
 
-To register the tasks supported by your library on the hub you'll need to add a mapping from your library name to its supported tasks in this [file](https://github.com/huggingface/huggingface.js/blob/main/packages/tasks/src/library-to-tasks.ts). This will ensure the inference API is registered for tasks supported by your model. This file is automatically generated as part of a [GitHub Action](https://github.com/huggingface/api-inference-community/actions/workflows/python-api-export-tasks.yaml) in the [
+To register the tasks supported by your library on the hub you'll need to add a mapping from your library name to its supported tasks in [library-to-tasks.ts file](https://github.com/huggingface/huggingface.js/blob/main/packages/tasks/src/library-to-tasks.ts). This will ensure the inference API is registered for tasks supported by your model. This file is automatically generated as part of a [GitHub Action](https://github.com/huggingface/api-inference-community/actions/workflows/python-api-export-tasks.yaml) in the [
```
pcuenca, line 187: I find this a bit confusing. It says you have to edit the file and then that it's automatically generated. The GitHub action generates [something similar](https://github.com/huggingface/api-inference-community/actions/runs/7017787981/job/19091817114), to the [`.ts` file](https://github.com/huggingface/huggingface.js/blob/main/packages/tasks/src/library-to-tasks.ts); for example, `audio-source-separation` is commented in the `.ts` file but not in the generated result.

My understanding is that the GitHub action provides the source of truth to the inference API, but I don't fully understand the workflow or relationship with `library-to-tasks.ts`.

pcuenca (commented):
Looks great! I don't fully understand the library-to-tasks workflow; not really the purpose of this PR but we may want to clarify?

PR #1151, opened 2023-12-01
Update instructions to add a library

coyotte508:
As reported in https://github.com/huggingface/hub-docs/pull/1150

PR #1150, opened 2023-12-01
Integrate SetFit library

tomaarsen:
Hello!

## Pull Request overview
* Integrate with the [SetFit](https://github.com/huggingface/setfit) library for Text Classification.

## Details
[SetFit](https://github.com/huggingface/setfit) is a library for text classification with ~1200 models on the Hub at the time of writing. A v1.0.0 release is upcoming, and it's a good time to add this widget support, etc. It can be used like so:
```python
from setfit import SetFitModel

model = SetFitModel.from_pretrained("tomaarsen/span-marker-bert-base-fewnerd-fine-super")
```
```
model.predict(["That was an awful movie"])
# => ["negative"]
```

I've followed the documentation [here](https://huggingface.co/docs/hub/models-adding-libraries), but it seems to be a bit outdated compared to the last time that I added an integration. It still mentions to edit `model-libraries.ts` with a snippet like
```js
const asteroid = (model: ModelData) =>
`from asteroid.models import BaseModel
  
model = BaseModel.from_pretrained("${model.id}")`;
```
But that type of data isn't in that file anymore. (cc: @coyotte508 ?)
Presumably I also have to update that in huggingface.js?

Note: Some of the links, e.g. https://huggingface.co/docs/setfit/reference/main#setfit.SetFitModel.predict, do not currently work, but will start working when [SetFit v1.0.0](https://github.com/huggingface/setfit/pull/439) releases next week. 

Related PRs:
* https://github.com/huggingface/api-inference-community/pull/359
* https://github.com/huggingface/huggingface.js/pull/387

- Tom Aarsen

PR #1152, opened 2023-12-01
Suggest adding lib to doc

coyotte508:
See https://github.com/huggingface/hub-docs/pull/1150 / comment in https://github.com/huggingface/hub-docs/pull/1151

Not sure at all where to add it, added at the end


osanseviero (approved):
Thank you!

PR #1167, opened 2023-12-19
Document `base_model` as a string (and optionally a list)

julien-c:


osanseviero (approved):

osanseviero:
Maybe also mention merges in https://huggingface.co/docs/hub/model-cards#specifying-a-base-model

julien-c:
> Maybe also mention merges in https://huggingface.co/docs/hub/model-cards#specifying-a-base-model

maybe in a subsequent PR cc @davanstrien?

PR #1177, opened 2024-01-05
no more need for the directory to be a split name

severo:
see https://github.com/huggingface/datasets/issues/6442#issuecomment-1822561211

I already fixed the example datasets (4, 5, 6): https://huggingface.co/collections/datasets-examples/image-dataset-6568e7cf28639db76eb92d65


lhoestq:
Actually I like the idea of keeping it to build the reader's intuition about the way they can define files and directories

severo:
OK. I'll revert the example datasets...

severo:
Done

PR #1197, opened 2024-01-23
Update docs for new Giskard Docker template on Spaces

luca-martial:
Clarified some info about token usage in the Space


osanseviero (approved):
Thanks!

PR #1200, opened 2024-01-25
Correct `info.json` example contents in Unity Sentis documentation.

nilq:
A bit of nitpick, but the example contents of `info.json` are valid JSON. They're missing quotation for the keys, and are also using ” (ASCII 8220) instead of ” (ASCII 34) for the string values.


pcuenca (approved):
You are right, @nilsq, thanks a lot!

PR #1209, opened 2024-02-01
Document space mini header

krampstudio:
See https://github.com/huggingface/moon-landing/pull/8753


julien-c (approved):
let's keep it a bit low-key / easter egg, but lgtm:)

(we could add a screenshot if we wanted too.. also that page starts to be quite long, maybe we'd want to split into sections at some point)

PR #1216, opened 2024-02-09
Mention the checkup period of two factor auth

Kakulukian:
I think it's clear?


docs/hub/security-two-fa.md
```diff
@@ -35,6 +35,10 @@ In Hugging Face Hub:
 
 If you entered the correct pin, the Hub displays a list of recovery codes. Download them and keep them in a safe place.
 
+<Tip>
+After configuring 2FA, your account enters a 30-day check-up period. To exit this period, successfully complete 2FA within those 30 days. If you do not complete 2FA during this time, you will be prompted to perform 2FA in an existing session on the 30th day.
```
SBrandeis, line 39: A bit convoluted IMO, maybe we can do something simpler

Would this be a correct assertion?
```suggestion
You will be prompted for 2FA on every new session, and every 30 days for active sessions.
```

Kakulukian: Yes it is

SBrandeis: "session" might not be the clearest wording here

coyotte508:
```suggestion
You will be prompted for 2FA every time you log in, and after that every 30 days.
```

coyotte508: or cc @Michellehbn x)

julien-c, line 39:
```suggestion
You will be prompted for 2FA every time you log in, and every 30 days.
```

julien-c (approved):
did a mix of the two formulations, LGTM like this

julien-c:
also please consider merging https://github.com/huggingface/hub-docs/pull/1217 too

PR #1227, opened 2024-02-28
✨ Document oauth duration for spaces

coyotte508:
I thought I already did it :facepalm:


julien-c (approved):
lgtm, will there also be a token refresh flow at some point? (maybe worth mentioning it?)

osanseviero (approved):
Thanks!

coyotte508:
> lgtm, will there also be a token refresh flow at some point? (maybe worth mentioning it?)

maybe :thinking:

PR #1239, opened 2024-03-12
Add section to explain which format to use for dataset files

### LINKED ISSUE #1232: Add doc about suggested formats for dataset files
Based on the forum/Discord threads, it's often unclear to users which format to use for dataset files uploaded to the Hub. So besides [listing](https://huggingface.co/docs/hub/datasets-adding#file-formats) the supported formats, it would also be nice to give more guidance on which one to choose.

cc @severo @lhoestq

mariosasko:
Close #1232

### DIFF
--- a/docs/hub/datasets-adding.md
+++ b/docs/hub/datasets-adding.md
@@ -96,7 +96,7 @@ Other formats and structures may not be recognized by the Hub.
 
 ### Which file format should I use?
 
-For most types of datasets, Parquet is the recommended format due to its efficient compression and fast read speed. Alternatively, CSV or JSON Lines/JSON can be used for non-nested tabular data. Although easy to parse, compared to Parquet, these formats require more space and are slower to read, so using them for data larger than several GBs is not recommended. Parquet is less efficient for storing images and audio, so uploading raw files is better in that case. For large scale image and audio datasets, [WebDataset](https://github.com/webdataset/webdataset) should be preferred over raw image and audio files to avoid the overhead of accessing individual files.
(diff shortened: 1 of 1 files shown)


severo (approved):
good addition!

docs/hub/datasets-adding.md
```diff
@@ -94,6 +94,10 @@ Image and audio files can also have additional metadata files, see the [Data fil
 You may want to convert your files to these formats to benefit from all the Hub features.
 Other formats and structures may not be recognized by the Hub.
 
+### Which file format should I use?
+
+For most types of datasets, Parquet is the recommended format due to its efficient compression and fast read speed. Alternatively, CSV or JSON Lines/JSON can be used for non-nested tabular data. Although easy to parse, compared to Parquet, these formats require more space and are slower to read, so using them for data larger than several GBs is not recommended. Parquet is less efficient for storing images and audio, so uploading raw files is better in that case. For large scale image and audio datasets, [WebDataset](https://github.com/webdataset/webdataset) should be preferred over raw image and audio files to avoid the overhead of accessing individual files.
```
lhoestq, line 99: I added a suggestion since technically csv/json are faster to read than Parquet since they're uncompressed (maybe you meant "faster to download"). Also JSON Lines is fine for nested data.

I also mentioned in which cace people should use Parquet over WebDataset for large scale image and audio datasets. Feel free to edit

```suggestion
For most types of datasets, Parquet is the recommended format due to its efficient compression, rich typing, and since a variety of tools supports this format with optimized read and batched operations. Alternatively, CSV or JSON Lines/JSON can be used for tabular data (prefer JSON Lines for nested data). Although easy to parse compared to Parquet, these formats are not recommended for data larger than several GBs. For image and audio datasets, uploading raw files is the most practical for most use cases since it's easy to access individual files. For large scale image and audio datasets streaming, [WebDataset](https://github.com/webdataset/webdataset) should be preferred over raw image and audio files to avoid the overhead of accessing individual files. Though for more general use cases involving analytics, data filtering or metadata parsing, Parquet is the recommended option for large scale image and audio datasets.
```

severo: These modifs help a lot, but I'm wondering if we could improve the structure a bit. For example, we start recommending Parquet, and we do it again in the last sentence. Maybe one paragraph per format, with its strengths and defects, to help parse the info quickly? feel free to ignore

mariosasko: I think this is a good start. And we can improve the structure later (when more info is added to this section).

lhoestq: Maybe a table with formats vs use case or something like that

PR #1240, opened 2024-03-12
Update spaces-more-ways-to-create.md

aroffe99:
Remove ']' so the text becomes link-able


julien-c (approved):
thanks for fixing @aroffe99!

PR #1241, opened 2024-03-13
✨ succinct documentation for resource groups

SBrandeis:
cc @Kakulukian @coyotte508 @Pierrci if I missed something or wrote anything wrong

Ping @Michellehbn for a native speaker review 🙏

PR #1245, opened 2024-03-14
Follow up to #1241

Pierrci:


julien-c (approved):
merged the previous one too soon it looks like:)

PR #1252, opened 2024-03-19
📝 Document webhooks.updatedRefs

coyotte508:
cc @Wauplin for viz


Wauplin (approved):
Thanks!

Just to be sure, does that mean `webhook.version` got updated from "3" to "4" or not? (in that case we should update it in the docs as well)

coyotte508:
> Thanks!
> 
> Just to be sure, does that mean `payload.webhook.version` got updated from "3" to "4" or not? (in that case we should update it in the docs as well)

Nope, will only change the version on breaking changes (which I hope there are none)

osanseviero (approved):
Nice!

pcuenca (approved):
Great work, clear and concise imo 🔥

julien-c (commented):
you should post about this publicly given it can unlock some cool use cases IMO

PR #1256, opened 2024-03-25
Add new gemma license

osanseviero:


pcuenca (approved):
Thank you!

julien-c (commented):
is this the id that they've been using in the repos?

osanseviero:
No, they are using `other`. We'll need to update their repos, currently:

```
license: other
license_name: gemma-terms-of-use
license_link: https://ai.google.dev/gemma/terms
```

julien-c:
yep i meant for `license_name` i.E. i would have used `gemma-terms-of-use` as the id here. No big deal though

PR #1267, opened 2024-04-10
Update model-card-annotated.md

meg-huggingface:
Food for thought: Adding a section that deals directly with Redteaming (as that is becoming a big interest/concern). Called it "Safety Testing", so it's not limited to redteaming and can also include adversarial testing (which some folks in redteaming world make a distinction about). Hit specifically on CSAM, as we have been inspired by Thorn and Rebecca Portnoff to be more proactive about this. For more context, see: https://huggingface.slack.com/archives/C0317KZSB5H/p1712707749145689?thread_ts=1712614250.080999&cid=C0317KZSB5H

### DIFF
--- a/docs/hub/model-card-annotated.md
+++ b/docs/hub/model-card-annotated.md
@@ -182,28 +182,30 @@ _Detail throughput, start/end time, checkpoint sizes, etc._
 
 # Evaluation
 
-**Section Overview:** This section describes the evaluation protocols, what is being measured in the evaluation, and provides the results.  Evaluation ideally has at least two parts, with one part looking at general performance ([Testing Data, Factors & Metrics](#testing-data-factors--metrics)) and another looking at performance with respect to specific social safety issues ([Societal Impact Assessment](#societal-impact-assessment-optional)), such as may be done with red-teaming. The former is ideally constructed with factors, such as domain and demographic subgroup, and metrics, such as accuracy, which are prioritized in light of foreseeable error contexts and groups. Target fairness metrics should be decided based on which errors are more likely to be problematic in light of the model use. You can also specify your model's evaluation results in a structured way in the model card metadata. Results are parsed by the Hub and displayed in a widget on the model page. See https://huggingface.co/docs/hub/model-cards#evaluation-results.
+**Section Overview:** This section describes the evaluation protocols, what is being measured in the evaluation, and provides the results.  Evaluation ideally has at least two parts, with one part looking at quantitative measurement of general performance ([Testing Data, Factors & Metrics](#testing-data-factors--metrics)), such as may be done with benchmarking; and another looking at performance with respect to specific social safety issues ([Societal Impact Assessment](#societal-impact-assessment-optional)), such as may be done with red-teaming. You can also specify your model's evaluation results in a structured way in the model card metadata. Results are parsed by the Hub and displayed in a widget on the model page. See https://huggingface.co/docs/hub/model-cards#evaluation-results.
 
 
 ## Testing Data, Factors & Metrics
+
+_Evaluation is ideally **disaggregated** with respect to different factors, such as task, domain and population subgroup; and calculated with metrics that are most meaningful for foreseeable contexts of use. Equal evaluation performance across different subgroups is said to be "fair" across those subgroups; target fairness metrics should be decided based on which errors are more likely to be problematic in light of the model use. However, this section is most commonly used to report aggregate evaluation performance on different task benchmarks._
 
 ### Testing Data
 
 `testing_data`
 
-_Ideally this links to a Dataset Card for the testing data._
+_Describe testing data or link to its Dataset Card._
 
 ### Factors
 
 `testing_factors`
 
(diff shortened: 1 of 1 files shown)


meg-huggingface:
Thanks @julien-c , completely agree it's a bit heavy and also agree we should keep model cards as light and as generally applicable as possible, appreciate that.
I have been waffling on best approaches with respect to 2 things happening at the moment:

1. An increased interest in "safety" and "redteaming" in AI, including in regulation, and yet no clearly corresponding place where one might put results from this in model cards. In "redteaming", topics of concern include sex, violence (and things that might cause violence), racism/sexism etc. So a place to report these types of things would be handy, although I don't necessarily want to be "redteam" specific, since there is the related strategy of using "adversarial testing", as well as approaches that calculate scores over outputs, such as tf-idf and nPMI. So creating a section that all these approaches are all getting at, and recognizing it is part of the evaluation protocol -- calling it something like a "Societal Impact" section -- might be handy to have. But note there is a relationship to "Bias & Risks" that we would need to be clear about. (I detail this more below.)
2. Thorn's advocacy for having companies start paying some serious attention to the harms that Thorn is focusing on, with a suggestion from them for updating the Model Card protocol in particular. After discussing in the E&S meeting, it's probably good for models **not** to be too specific about some of these issues, e.g., with searchable tags,  in part because this enables malicious actors to *find* models with potential harm more easily.

To get to your point about applicability to different models: Although the interest in this has grown for generative models, the kinds of issues are applicable to different kinds of models:
- Classification models: Creating inappropriate "tags" corresponding to violence or sex. For example, tagging a thermometer gun with a black hand as a firearm: https://algorithmwatch.org/en/google-vision-racism/ . This is the type of thing that would be caught in "redteaming".
- Models based on tabular data: Can be things like regression. Forecasting, health, generally filling in a cell for unseen data. As with all models, the potential harm really depends on the data. But relevant problematic outputs could have to do with protecting peoples' PRIVACY (does sensitive private data get leaked from the model?) or for example in the medical domain, drug recommendations that would lead to seizures or be inappropriate for children, etc.

So, in light of all of the above:
What about creating a section in the Testing/Evaluation part that said "Societal Impact" or "Safety", was **optional**, and was free text? Then in the Annotated Model Card, we could explain some of the issues that people might check for (privacy, violence, sexualization), but explicitly **not** have tags for them to make it less "heavy" and circumvent the foreseeable issue were malicious actors search for models that are harmful specifically?
The distinction with "Bias, Risks, and Limitations" is that people tend to put very general things there "may produce incorrect information, may be biased wrt skin tone" etc., whereas the "Societal Impact"/"Safety" section would speak directly to what people are now testing for.

julien-c:
yep, sounds like a good trade off to me!

pcuenca:
> "Societal Impact" or "Safety"

How about `Safety Assessment`, so the goal is to encourage the reporting of specific evaluations the model authors may have undertaken? "Societal Impact" might be more accurate, but for a casual reader (me) the distinction with respect to "Bias, Risks, and Limitations" might be a bit blurred. "Safety" may be more generic if we expect authors to also include information about guardrails or similar mechanisms.

meg-huggingface:
> > "Societal Impact" or "Safety"
> 
> How about `Safety Assessment`, so the goal is to encourage the reporting of specific evaluations the model authors may have undertaken? "Societal Impact" might be more accurate, but for a casual reader (me) the distinction with respect to "Bias, Risks, and Limitations" might be a bit blurred. "Safety" may be more generic if we expect authors to also include information about guardrails or similar mechanisms.

That makes a lot of sense to me!

docs/hub/model-card-annotated.md
```diff
@@ -182,7 +182,7 @@ _Detail throughput, start/end time, checkpoint sizes, etc._
...
+**Section Overview:** This section describes the evaluation protocols, what is being measured in the evaluation, and provides the results.  Evaluation ideally has at least two parts, with one part looking at general performance ([Testing Data, Factors & Metrics](#testing-data-factors--metrics)) and another looking at performance with respect to specific social safety issues ([Societal Impact Assessment](#societal-impact-assessment-optional)), such as may be done with red-teaming. The former is ideally constructed with factors, such as domain and demographic subgroup, and metrics, such as accuracy, which are prioritized in light of foreseeable error contexts and groups. Target fairness metrics should be decided based on which errors are more likely to be problematic in light of the model use. You can also specify your model's evaluation results in a structured way in the model card metadata. Results are parsed by the Hub and displayed in a widget on the model page. See https://huggingface.co/docs/hub/model-cards#evaluation-results.
```
pcuenca, line 185: Everything after `The former` only applies to the first group, including the structured reporting of `evaluation results`, which could be interpreted as being comprehensive of societal impact / red-teaming results (while reading it I was trying to parse what would be applicable to the Societal Impact Assessment topic). We could add a final comment about the Societal Impact Assessment being free text, but it's probably not necessary as it's explained later. Maybe use `quantitative evaluation results`?

pcuenca (approved):

docs/hub/model-card-annotated.md
meg-huggingface: Gotcha, yes. What about moving the instructions for each into each corresponding section? Let me give it a whirl.

pcuenca (approved):

PR #1278, opened 2024-04-28
Add quotes so that a git command works on fish shell too

osbm:
This is a pretty small change. But it could make the page a little more smooth.

This new quoted version works in both bash and fish.


julien-c (approved):
ok for me, thanks

Issue #1222, opened 2024-02-21
Link to datasets YAML configuration page


severo:
Link to https://huggingface.co/docs/datasets/v2.7.1/en/dataset_card#more-yaml-tags from https://huggingface.co/docs/hub/datasets-manual-configuration, to complement with all the possible values in README's YAML

severo:
And give an example of each supported feature type in the YAML config. See https://discuss.huggingface.co/t/appropriate-yaml-for-dataset-info-list-float/74418 for example: I think we currently have no reference to share to the user.

severo:
Indeed, it has been removed in https://github.com/huggingface/datasets/pull/5470#discussion_r1088471903

The spec is here: https://github.com/huggingface/hub-docs/blob/main/datasetcard.md?plain=1

severo:
Somewhat related: discussion about the spec: https://github.com/huggingface/dataset-viewer/issues/2639

Also: should we just redirect to the spec (https://github.com/huggingface/hub-docs/blob/main/datasetcard.md?plain=1), or should we create a dedicated doc page for this? Adding the link would already by a good step forward.

severo:
Also: https://github.com/huggingface/hub-docs/blob/main/datasetcard.md?plain=1 is outdated:

```yaml
configs:  # Optional for datasets with multiple configurations like glue.
- {config_0}  # Example for glue: sst2
- {config_1}  # Example for glue: cola
```

It does not respect the current format: https://huggingface.co/docs/hub/datasets-manual-configuration.

Ideally, it should be the reference, with more details than https://huggingface.co/docs/hub/datasets-manual-configuration, not the other way.

cc @polinaeterna for example if you want to look at it

lappemic:
> Adding the link would already by a good step forward.

Shall i start out with this and have a look where it leads us @severo? Or would you suggest a different approachch?

severo:
Hmmm, I think we have to improve the spec first. Then, link to it from the docs page, otherwise the link would not bring much value.

severo:
Do you want to work on a PR to improve the spec https://github.com/huggingface/hub-docs/blob/main/datasetcard.md?plain=1 ?
The idea is to add the structure of the `configs:` field, to match https://huggingface.co/docs/hub/datasets-manual-configuration at least (config_name, data_files, etc). Some more fields can be passed, if I'm not wrong (it's defined in https://github.com/huggingface/datasets, but @polinaeterna knows these details better than I)

[...]

PR #1292, opened 2024-05-17
add config_name and data_files to datasetcard.md specs

lappemic:
As realised in #1222 the specs of the datasetcard are not up to date. This PR therefore updates the specs.

Let me know if this is sufficient or you would like to add more config examples.

Cc: @severo, @polinaeterna

### DIFF
--- a/datasetcard.md
+++ b/datasetcard.md
@@ -33,13 +33,15 @@ task_ids:
 - {subtask_0}  # Example: extractive-qa
 - {subtask_1}  # Example: multi-class-image-classification
 paperswithcode_id: {paperswithcode_id}  # Dataset id on PapersWithCode (from the URL). Example for SQuAD: squad
-configs:  # Optional for datasets with multiple configurations.
-  - config_name: default
+configs:  # Optional. This can be used to pass additional parameters to the dataset loader, such as `data_files`, `data_dir`, and any builder-specific parameters  
+  - config_name: {config_name_0}  # Example: default
     data_files:
-      - split: train
-        path: "data.csv"
-      - split: test
-        path: "holdout.csv"
+      - split: {split_name_0}  # Example: train
+        path: {file_path_0}  # Example: data.csv
+      - split: {split_name_1}  # Example: test
+        path: {file_path_1}   # Example: holdout.csv
+   - config_name: {config_name_1}  # Example: ...
+      ... #TODO
 
 # Optional. This part can be used to store the feature types and size of the dataset to be used in python. This can be automatically generated using the datasets-cli.
 dataset_info:
--- a/datasetcard.md
+++ b/datasetcard.md
@@ -41,7 +41,6 @@ configs:  # Optional. This can be used to pass additional parameters to the data
   - split: {split_name_1}  # Example: test
     path: {file_path_1}   # Example: holdout.csv
 - config_name: {config_name_1}  # Example: ...
-  ... #TODO
 
 # Optional. This part can be used to store the feature types and size of the dataset to be used in python. This can be automatically generated using the datasets-cli.
 dataset_info:
--- a/datasetcard.md
+++ b/datasetcard.md
@@ -34,14 +34,14 @@ task_ids:
 - {subtask_1}  # Example: multi-class-image-classification
 paperswithcode_id: {paperswithcode_id}  # Dataset id on PapersWithCode (from the URL). Example for SQuAD: squad
 configs:  # Optional. This can be used to pass additional parameters to the dataset loader, such as `data_files`, `data_dir`, and any builder-specific parameters  
-  - config_name: {config_name_0}  # Example: default
-    data_files:
-      - split: {split_name_0}  # Example: train
-        path: {file_path_0}  # Example: data.csv
-      - split: {split_name_1}  # Example: test
-        path: {file_path_1}   # Example: holdout.csv
-   - config_name: {config_name_1}  # Example: ...
-      ... #TODO
+- config_name: {config_name_0}  # Example: default
+  data_files:
+  - split: {split_name_0}  # Example: train
+    path: {file_path_0}  # Example: data.csv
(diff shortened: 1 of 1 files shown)


julien-c (approved):
LGTM, thanks

Note that there's also a `config_names: list[str]` (i.e the previous property) but i don't think we need to document it

datasetcard.md
```diff
@@ -33,9 +33,13 @@ task_ids:
...
-configs:  # Optional for datasets with multiple configurations like glue.
-- {config_0}  # Example for glue: sst2
-- {config_1}  # Example for glue: cola
+configs:  # Optional for datasets with multiple configurations.
+  - config_name: default
+    data_files:
+      - split: train
+        path: "data.csv"
+      - split: test
+        path: "holdout.csv"
```
polinaeterna, line 42:
```suggestion
configs:  # Optional. This can be used to pass additional parameters to the dataset loader, such as `data_files`, `data_dir`, and any builder-specific parameters  
  - config_name: {config_name_0}  # Example: default
    data_files:
      - split: {split_name_0}  # Example: train
        path: {file_path_0}  # Example: data.csv
      - split: {split_name_1}  # Example: test
        path: {file_path_1}   # Example: holdout.csv
   - config_name: {config_name_1}  # Example: ...
      ... #TODO
```
and then we can also add example, ideally for some real dataset with multiple configs, maybe copy a part from [nyu-mll/glue](https://huggingface.co/datasets/nyu-mll/glue/blob/main/README.md?code=true#L346)?

polinaeterna (changes requested):
thank you a lot for working on this! 
I've left a suggestion to reformat it a bit to keep it aligned with the rest of the file.

Also I wonder if we should try to include all possible cases in these specs or keep it simple as it is now and refer to https://huggingface.co/docs/hub/datasets-manual-configuration instead? cc @severo  because there are a lot of parameters apart from `data_files` that can be passed, for example for `csv` builder one can specify a `sep`, for audio/imagefolder builders - `drop_metadata`/`drop_labels`, etc. there is also `data_dir` parameter, similar to `data_files`.
I personally would keep it simple but not sure cc @lhoestq @stevhliu

datasetcard.md
lappemic: Looks clearer, thanks. 

About the example: How about using [ELI5 dataset card](https://huggingface.co/datasets/eli5/blob/main/README.md)? I come up because it is already used in [Dataset card creatino guide](https://huggingface.co/docs/hub/datasets-cards#dataset-card-creation-guide). I find it always easier when the same example is referenced. WDYT?

stevhliu:
> I personally would keep it simple but not sure

Since this is just an example dataset card, we don't necessarily need to include all possible parameters. A simple example use case should suffice, with a link to more options (to allow users to explore more gradually if they'd like), as you suggested sounds great to me.

datasetcard.md
```diff
@@ -33,9 +33,15 @@ task_ids:
...
-- {config_1}  # Example for glue: cola
+configs:  # Optional. This can be used to pass additional parameters to the dataset loader, such as `data_files`, `data_dir`, and any builder-specific parameters  
+  - config_name: {config_name_0}  # Example: default
+    data_files:
+      - split: {split_name_0}  # Example: train
+        path: {file_path_0}  # Example: data.csv
+      - split: {split_name_1}  # Example: test
+        path: {file_path_1}   # Example: holdout.csv
+   - config_name: {config_name_1}  # Example: ...
+      ... #TODO
```
lhoestq, line 44: (nit) smaller indentation
```suggestion
configs:  # Optional. This can be used to pass additional parameters to the dataset loader, such as `data_files`, `data_dir`, and any builder-specific parameters  
- config_name: {config_name_0}  # Example: default
  data_files:
  - split: {split_name_0}  # Example: train
    path: {file_path_0}  # Example: data.csv
  - split: {split_name_1}  # Example: test
    path: {file_path_1}   # Example: holdout.csv
- config_name: {config_name_1}  # Example: ...
  ... #TODO
```

lhoestq (approved):
thanks for the addition ! looks like we missed this update

> Also I wonder if we should try to include all possible cases in these specs or keep it simple as it is now and refer to https://huggingface.co/docs/hub/datasets-manual-configuration instead? cc @severo because there are a lot of parameters apart from data_files that can be passed, for example for csv builder one can specify a sep, for audio/imagefolder builders - drop_metadata/drop_labels, etc. there is also data_dir parameter, similar to data_files.
I personally would keep it simple but not sure cc @lhoestq @stevhliu

I think it's fine for now. We only have this [page](https://huggingface.co/docs/datasets/en/package_reference/loading_methods#from-files) for a full list of parameters, it would be cool to have a more readable page for this that we could link to from this spec file later

datasetcard.md
julien-c: yes, and remove the #TODO at the end

datasetcard.md
```diff
@@ -33,9 +33,14 @@ task_ids:
...
-- {config_0}  # Example for glue: sst2
-- {config_1}  # Example for glue: cola
+configs:  # Optional. This can be used to pass additional parameters to the dataset loader, such as `data_files`, `data_dir`, and any builder-specific parameters  
+- config_name: {config_name_0}  # Example: default
+  data_files:
+  - split: {split_name_0}  # Example: train
+    path: {file_path_0}  # Example: data.csv
+  - split: {split_name_1}  # Example: test
+    path: {file_path_1}   # Example: holdout.csv
+- config_name: {config_name_1}  # Example: ...
```
polinaeterna, line 43:
```suggestion
- config_name: {config_name_1}  # Example: processed
  data_files:
    - split: {split_name_3}  # Example: train
      path: {file_path_3}  # Example: data_processed.csv
```
let's complete this block for second config? and we can merge

polinaeterna (approved):

PR #1303, opened 2024-06-02
Fix invalid Python syntax for Spaces Overview page

alexandercarruthers:
`print(os.getenv['MODEL_REPO_ID'])` is invalid Python syntax this has been replaced with `print(os.getenv('MODEL_REPO_ID'))`


osanseviero (approved):
Thanks!

PR #1304, opened 2024-06-03
remove the mention given the app hasn't been maintained AFAIK

julien-c:


Wauplin (commented):
> given the app hasn't been maintained

Or should we maintain it? Space still seems to be working. It has received 1470 visits on the last month

(we can remove the `New !` though)

osanseviero (approved):

julien-c:
@Wauplin couple of pages are broken: 
image

for me in the context of the doc page https://huggingface.co/docs/hub/model-cards the app is not super useful

Note that i'm not proposing of removing or deprecating the Space, just the mention on that doc page (or we can put it somewhere else on the page, less visible?)

Wauplin:
Fine for me then :)

meg-huggingface:
Yes, I agree as well. Sounds good to me.

meg-huggingface (approved):
SGTM

PR #1306, opened 2024-06-06
Add a section about email status for enterprise org

Kakulukian:


docs/hub/organizations-security.md
```diff
@@ -31,3 +31,14 @@ As an organization `admin`, go to the **Members** section of the org settings to
 <img class="hidden dark:block" src="https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/hub/org-members-page-dark.png"/>
 </div>
 
+
+## Viewing members' email address
+
+<Tip warning={true}>
+This feature is part of the <a href="https://huggingface.co/enterprise" target="_blank">Enterprise Hub</a>.
+</Tip>
```
julien-c, line 39: need a screenshot that's the most improtant part imo

Kakulukian: wdym?

julien-c: can we have something like this, but where we can see some (obfuscated or dummy) email addresses (and the other columns too)? 

image

SBrandeis (approved):
Thank you ❤️

PR #1310, opened 2024-06-18
Document paper endpoints

osanseviero:


docs/hub/api.md
```diff
@@ -258,6 +259,18 @@ headers = { "authorization" :  "Bearer $token" }
 
 This is equivalent to `huggingface_hub.whoami()`.
 
+## Paper Pages API
+
+The following endpoint gets information about a paper.
+
+### GET /api/arxiv/{arxiv_id}/repos
+
+Get all the models, datasets, and Spaces mentioned in a paper.
```
julien-c, line 268: we won't be able to use the hits for this to represent actual calls from Arxiv 

but i guess that's fine?

osanseviero: From my side it's fine

julien-c: BTW this is not really what this is doing

```suggestion
### GET /api/arxiv/{arxiv_id}/repos

Get all the models, datasets, and Spaces that refer to a paper.
```

better?^

julien-c (approved):
ok for me with my suggestion

SBrandeis (approved):

PR #1322, opened 2024-06-29
Update Livebook setup instructions

jonatanklosko:
We tell users to set `XLA_TARGET` which tells the Elixir's tensor library to use the GPU. We changed the default Docker image for the space to use CUDA 12, so this updates the docs accordingly.

### DIFF
--- a/docs/hub/spaces-sdks-docker-livebook.md
+++ b/docs/hub/spaces-sdks-docker-livebook.md
@@ -60,7 +60,7 @@ Here's a quick video showing how to do that:
 
 If your Space Hardware is configured to use a GPU accelerator, you can configure Livebook to leverage it.
 
-Go to the Settings page of your Space and create a secret called `XLA_TARGET` with the value `cuda118`.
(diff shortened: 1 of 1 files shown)


docs/hub/spaces-sdks-docker-livebook.md
```diff
@@ -60,7 +60,7 @@ Here's a quick video showing how to do that:
 
 If your Space Hardware is configured to use a GPU accelerator, you can configure Livebook to leverage it.
 
-Go to the Settings page of your Space and create a secret called `XLA_TARGET` with the value `cuda118`.
+Go to the Settings page of your Space and create a secret called `XLA_TARGET` with the value `cuda12` (this value may differ, depending on the Nx/EXLA versions you use, see [elixir-nx/xla](https://github.com/elixir-nx/xla?tab=readme-ov-file#xla_target) for the latest reference).
```
jonatanklosko, line 63: Actually, I've just realised we that particular change is not released yet. So I will just update to cuda 12, since that matches the default image we now use. I will update this docs again once the change is released.

```suggestion
Go to the Settings page of your Space and create a secret called `XLA_TARGET` with the value `cuda120` (this value may differ, depending on the Nx/EXLA versions you use, see [elixir-nx/xla](https://github.com/elixir-nx/xla?tab=readme-ov-file#xla_target) for the latest reference).
```

jonatanklosko:
Actually, I've just realised that it's probably best for us to set that env var in the Dockerfile directly. I will open a new PR later!

PR #1324, opened 2024-07-01
Update Livebook instructions

jonatanklosko:
We now set the env var in the Dockerfile itself, so GPU should be used by default when available, and in the future the env var won't be needed altogether.


osanseviero (approved):
Nice 🔥

PR #1326, opened 2024-07-03
Update llama.cpp docs

Vaibhavs10:


docs/hub/gguf-llamacpp.md
```diff
@@ -1,16 +1,46 @@
 # GGUF usage with llama.cpp
 
-Llama.cpp directly allows you to download and run inference on a GGUF simply by providing a path to the Hugging Face repo path and the file name. llama.cpp would download the model checkpoint in the directory you invoke it from:
+Llama.cpp allows you to download and run inference on a GGUF simply by providing a path to the Hugging Face repo path and the file name. llama.cpp would download the model checkpoint in the directory you invoke it from:
```
cakiki, line 3: r/would/will?

docs/hub/gguf-llamacpp.md
```diff
@@ -1,16 +1,46 @@
...
+Additionally, you can invoke an OpenAI spec chat completions endpoint directly using the llama.cpp server
+
+```bash
+./llama-server \
+  --hf-repo lmstudio-community/Meta-Llama-3-8B-Instruct-GGUF \
+  -m Meta-Llama-3-8B-Instruct-Q8_0.gguf \
+  -p "I believe the meaning of life is " -n 128
+```
+
+After running the server you can simply utilise the endpoint as below:
```
cakiki, line 21: Are there guidelines about US vs UK spelling? (re:utilise vs utilize)

cakiki: Alternative: call the endpoint

Vaibhavs10: I follow UK spelling 🏰 :p

julien-c (commented):
very cool update!

docs/hub/gguf-llamacpp.md
```diff
@@ -1,16 +1,46 @@
...
+
+```
```
ngxson, line 23:
```suggestion
```bash
```

Nitpick, but `"model": "llama-3-8b",` is not necessary

docs/hub/gguf-llamacpp.md
```diff
@@ -1,16 +1,46 @@
 # GGUF usage with llama.cpp
 
-Llama.cpp directly allows you to download and run inference on a GGUF simply by providing a path to the Hugging Face repo path and the file name. llama.cpp would download the model checkpoint in the directory you invoke it from:
+Llama.cpp allows you to download and run inference on a GGUF simply by providing a path to the Hugging Face repo path and the file name. llama.cpp would download the model checkpoint in the directory you invoke it from:
 
 ```bash
-./main \
+./llama-cli \
```
ngxson, line 6: Should we also add a small instruction for conversation mode `-cnv` ?

I think most users will prefer to use chat models.

docs/hub/gguf-llamacpp.md
```diff
@@ -1,16 +1,46 @@
...
   -p "I believe the meaning of life is " -n 128
 ```
 
+Additionally, you can invoke an OpenAI spec chat completions endpoint directly using the llama.cpp server
+
+```bash
+./llama-server \
+  --hf-repo lmstudio-community/Meta-Llama-3-8B-Instruct-GGUF \
+  -m Meta-Llama-3-8B-Instruct-Q8_0.gguf \
+  -p "I believe the meaning of life is " -n 128
```
ngxson, line 18: no need for `-p` and `-n` for llama-server, since it can be controlled in the HTTP request.

The -p prompt is the system prompt, and -n number of tokens is mapped to `"max_tokens"` (see [OpenAI docs](https://platform.openai.com/docs/api-reference/chat/create))

docs/hub/gguf-llamacpp.md
```diff
@@ -1,16 +1,46 @@
...
+}'
+```
+
 Replace `--hf-repo` with any valid Hugging Face hub repo name and `-m` with the GGUF file name in the hub repo - off you go! 🦙
 
 Find more information [here](https://github.com/ggerganov/llama.cpp/pull/6234).
 
-Note: Remember to `build` llama.cpp with `LLAMA_CURL=ON` :)
\ No newline at end of file
+Note: Remember to `build` llama.cpp with `LLAMA_CURL=ON` :)
```
ngxson, line 46: Nitpick, but should we use the same as on the snippet on local apps: `LLAMA_CURL=1` ?

It does not technically change anything, but just in case some users get confused whether it should be ON or 1

docs/hub/gguf-llamacpp.md
ngxson, line 3: Should we also document where downloaded file (via `--hf-repo`) is cached? Just in case user wants to clear the cache. Ref: https://github.com/ggerganov/llama.cpp/pull/7826

docs/hub/gguf-llamacpp.md
```diff
@@ -1,16 +1,46 @@
...
   -m Meta-Llama-3-8B-Instruct-Q8_0.gguf \
   -p "I believe the meaning of life is " -n 128
 ```
 
+Additionally, you can invoke an OpenAI spec chat completions endpoint directly using the llama.cpp server
+
+```bash
+./llama-server \
+  --hf-repo lmstudio-community/Meta-Llama-3-8B-Instruct-GGUF \
+  -m Meta-Llama-3-8B-Instruct-Q8_0.gguf \
```
ngxson, line 17: `--hf-repo` should go with `--hf-file` (`-m` is technically the same, but `--hf-file` is to be more aligned with the snippet in local apps)

docs/hub/gguf-llamacpp.md
osanseviero, line 3:
```suggestion
Llama.cpp allows you to download and run inference on a GGUF simply by providing a path to the Hugging Face repo path and the file name. llama.cpp downloads the model checkpoint in the directory you invoke it from:
```

osanseviero (commented):
Looking nice 🔥

docs/hub/gguf-llamacpp.md
ngxson: (I re-read this, `llama.cpp would download the model checkpoint in the directory you invoke it from` ==> no longer true, because we now have the cache directory, as mentioned in my last comment)

docs/hub/gguf-llamacpp.md
Vaibhavs10: Good idea, removing it!

docs/hub/gguf-llamacpp.md
Vaibhavs10: Fixed in the latest commit.

docs/hub/gguf-llamacpp.md
Vaibhavs10: Good idea! Adding it in the snippet.

docs/hub/gguf-llamacpp.md
Vaibhavs10: Good call! I was lazy and just copied the binaries lol.

docs/hub/gguf-llamacpp.md
Vaibhavs10: Fixed!

docs/hub/gguf-llamacpp.md
Vaibhavs10: Updated.

julien-c (approved):
🔥

docs/hub/gguf-llamacpp.md
ngxson: I mean, the snippet should be split into 2 versions, because `-p` is used differently in 2 cases:

### Basic usage: run in conversation mode

```bash
./llama-cli
  --hf-repo lmstudio-community/Meta-Llama-3-8B-Instruct-GGUF \
  --hf-file Meta-Llama-3-8B-Instruct-Q8_0.gguf \
  -p "You are a helpful assistant" -cnv
```

### Advanced usage: run in completion mode

```bash
./llama-cli
  --hf-repo lmstudio-community/Meta-Llama-3-8B-Instruct-GGUF \
  --hf-file Meta-Llama-3-8B-Instruct-Q8_0.gguf \
  -p "I believe the meaning of life is " -n 128
```

Vaibhavs10: hmm. I think it might make it a bit too verbose for the doc. I can add a line just below it stating that you can run in `chat completion` mode, let me add it.

ngxson: Or maybe you can add only my first snippet?

When using `-cnv`, `-p` becomes system message, so it should be something like "You are helpful assistant" or "You are Bob". `-n` is also not relevant anymore in cnv mode, since the model will generate until it hits end-of-turn token.

Vaibhavs10: Fixed!

julien-c (commented):
Much improved doc page! https://huggingface.co/docs/hub/gguf-llamacpp

thx 🔥

PR #1346, opened 2024-07-22
Update datasets-fiftyone.md

severo:
more factual


lhoestq (approved):
sounds good, it's more aligned with the style of the other doc pages like that

PR #1350, opened 2024-07-25
add masader

severo:


lhoestq (approved):
Do you plan to add more ? xD

severo:
It does not depend on me :) I think this one is at least as important as the others, so I wanted to have it. Depending on how it evolve, we can rework it or remove later.

PR #1358, opened 2024-07-30
tiny nit while reading the `models-uploading` doc page

julien-c:


NielsRogge (approved):
Thanks, would you also be able to update line 20 from "#upload-a-pytorch-model-using-huggingface_hub" to "#upload-a-pytorch-model-using-huggingfacehub"?

Only the latter seems to correct link to the subsection.

Wauplin (approved):

julien-c:
@NielsRogge ok, 0479ef9d039422cf9022986318a421c2fb30fbdb

PR #1365, opened 2024-08-07
Model Library detection has changed on August 2nd

julien-c:


SBrandeis (approved):

pcuenca (approved):
Oh, St. Julien came! 😂

Vaibhavs10 (approved):
The day has finally come! 🔥 

Obligatory meme:
image

osanseviero (approved):
Thanks!

docs/hub/model-cards.md
```diff
@@ -98,10 +98,10 @@ tags:
 - flair
 ```
 
-If it's not specified, the Hub will try to automatically detect the library type. Unless your model is from `transformers`, this approach is discouraged and repo creators should use the explicit `library_name` as much as possible. 
+If it's not specified, the Hub will try to automatically detect the library type. However, this approach is discouraged and repo creators should use the explicit `library_name` as much as possible.
```
osanseviero, line 101:
```suggestion
If it's not specified, the Hub will try to automatically detect the library type. However, this approach is discouraged, and repo creators should use the explicit `library_name` as much as possible. 
```

julien-c:
thx for merging @osanseviero

PR #1366, opened 2024-08-08
Update spaces-sdks-static.md

Bluezeeofficial:


julien-c:
seems invalid/unwanted PR!

PR #1377, opened 2024-08-16
Fix links

severo:
Found with

```
muffet --rate-limit 5 -i "https://huggingface.co/docs/hub/*" https://huggingface.co/docs/hub/
```

### DIFF
--- a/docs/hub/datasets-adding.md
+++ b/docs/hub/datasets-adding.md
@@ -67,7 +67,7 @@ The rich features set in the `huggingface_hub` library allows you to manage repo
 
 ## Using other libraries
 
-Some libraries like [🤗 Datasets](../datasets/index), [Pandas](https://pandas.pydata.org/), [Polars](https://pola.rs), [Dask](https://www.dask.org/) or [DuckDB](https://duckdb.org/) can upload files to the Hub.
+Some libraries like [🤗 Datasets](https://huggingface.co/docs/datasets/index), [Pandas](https://pandas.pydata.org/), [Polars](https://pola.rs), [Dask](https://www.dask.org/) or [DuckDB](https://duckdb.org/) can upload files to the Hub.
 See the list of [Libraries supported by the Datasets Hub](./datasets-libraries) for more information.
 
 ## Using Git
--- a/docs/hub/datasets-adding.md
+++ b/docs/hub/datasets-adding.md
@@ -63,7 +63,7 @@ Adding a Dataset card is super valuable for helping users find your dataset and 
 
 ## Using the `huggingface_hub` client library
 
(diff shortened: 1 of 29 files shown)


julien-c (commented):
let's wait for @mishig25 review though because i think those links used to work

severo:
The detail: 
- in https://huggingface.co/docs/hub/datasets-dask, the link "Login with your Hugging Face account" works: https://huggingface.co/docs/huggingface_hub/quick-start#login
- but in https://huggingface.co/docs/hub/en/datasets-dask, it becomes https://huggingface.co/docs/hub/huggingface_hub/quick-start which is a 404

mishig25 (approved):
@severo thanks a lot for finding this problem.

Explanation why this problem is occurring.
1. We've used relative links such as `[🤗 Datasets](../datasets/index)` to switch from hub-docs page to another library's page.
2. This relative path works fine when there is no `{lang}` or `{branch}` path present in the URL.

Example:
`[🤗 Datasets](../datasets/index)` was going to different paths depending whether `{lang}` was present or not:
* /docs/hub/datasets-dask -> /docs/datasets/index **(good)**
* /docs/hub/en/datasets-dask -> /docs/hub/datasets/index **(bad)**

To clarify, using relative links is not a problem when you are linking pages within a same library (since they will share same `{lang}` or `{branch}` prefixes). Using relative links is a problem when linking pages from different libraries. Therefore, we should always use absolute path `/docs/{lib}/{page}` when linking pages from different libraries.

docs/hub/datasets-adding.md
```diff
@@ -67,7 +67,7 @@ The rich features set in the `huggingface_hub` library allows you to manage repo
 
 ## Using other libraries
 
-Some libraries like [🤗 Datasets](../datasets/index), [Pandas](https://pandas.pydata.org/), [Polars](https://pola.rs), [Dask](https://www.dask.org/) or [DuckDB](https://duckdb.org/) can upload files to the Hub.
+Some libraries like [🤗 Datasets](https://huggingface.co/docs/datasets/index), [Pandas](https://pandas.pydata.org/), [Polars](https://pola.rs), [Dask](https://www.dask.org/) or [DuckDB](https://duckdb.org/) can upload files to the Hub.
```
mishig25, line 70:
```suggestion
Some libraries like [🤗 Datasets](/docs/datasets/index), [Pandas](https://pandas.pydata.org/), [Polars](https://pola.rs), [Dask](https://www.dask.org/) or [DuckDB](https://duckdb.org/) can upload files to the Hub.
```

small nit: we can just do `/docs/**`

severo: OK, better! I'll change

mishig25: yep should apply to other diffs as well :)

severo:
Done. I changed `https://huggingface.co/docs` for `/docs` (but only in the datasets doc pages... I let the rest of the docs untouched)

pcuenca:
Thanks for the clarification @mishig25!

PR #1360, opened 2024-08-02
Add 'Audio dataset' doc page

severo:
fixes #1106 

Please also review the new collection: https://huggingface.co/collections/datasets-examples/audio-dataset-66aca0b73e8f69e3d069e607

---

Note: I could not create a Parquet example with the 4 example files. A possible idea would be to showcase how to have two audio columns:

```csv
sentence,as_a_cat,as_a_dog
"I'm hungry", <1.wav>, <3.wav>
"How are you?", <2.wav>, <4.wav>
```

Not mandatory for this PR, it can be added later, maybe with detailed code (not necessarily from the datasets library, by the way, maybe with duckdb or polars)

PR #1381, opened 2024-08-19
add audio dataset docs page to TOC

severo:
missing in #1360


albertvillanova (approved):
Thanks.

Just a nit: what about swapping Audio and Image, so they appear in alphabetical order?

PR #1390, opened 2024-08-26
fix: grammar in security-tokens doc

andre15silva:


merveenoyan (approved):
thanks!

julien-c:
this changes the meaning a bit though (i think)

andre15silva:
> this changes the meaning a bit though (i think)

It does!

Before the change, you recommend using fine-grained tokens only in production cases. This means that you do not recommend using fine-grained tokens in dev, for example.

After the change, you recommend only using fine-grained tokens in production cases. You do not make any recommendations for other scenarios.

PR #1417, opened 2024-09-16
Set context length in supported-models.md to the length that we effectively serve in API

aymeric-roucher:
Since we do not serve Llama-3.1 models up to their nominal context length of 128k, I think we should specify the context length we effectively serve so as not to create false expectations for the users.


osanseviero (approved):
LGTM, but maybe let's change the column name to "Supported Context Length". WDYT?

PR #1418, opened 2024-09-16
Update the Model Card Social Impact Assessment 

yjernite:
Proposing a rephrasing to move away from the binary "can the model to X" to "were risks pertaining to X evaluated"

For example for NCII, most image generation models that can generate some level of partial nudity can generate images of some quality that may qualify, but concerns of misuse will be drastically different depending on how easy it is to achieve those results, how realistic they look, etc.


julien-c (approved):
nice

osanseviero (approved):

PR #1422, opened 2024-09-21
Update pip installation of spacy models from hub spacy.md

Huertas97:
The update in the way to install spacy models from the hub comes from how pip parse the URL to check pacage name and version. As it detect's "any" as version, it can raise an error. Adding the name of the package alongside the URL solve this.  More info in this GitHub issue: https://github.com/explosion/spaCy/issues/13599

### DIFF
--- a/docs/hub/spacy.md
+++ b/docs/hub/spacy.md
@@ -24,7 +24,7 @@ All models on the Hub come up with useful features
 All `spaCy` models from the Hub can be directly installed using pip install.
 
 ```bash
-pip install https://huggingface.co/spacy/en_core_web_sm/resolve/main/en_core_web_sm-any-py3-none-any.whl
+pip install "en_core_web_sm @ https://huggingface.co/spacy/en_core_web_sm/resolve/main/en_core_web_sm-any-py3-none-any.whl"
 ```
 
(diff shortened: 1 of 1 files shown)


julien-c (approved):
LGTM, does this also work with older versions of pip?

Vaibhavs10 (approved):
Confirmed that this works on Python version 3.8 as well (repro below):

1. `uv venv --python 3.8`
2. `source .venv/bin/activate`
3. `pip install "en_core_web_sm @ https://huggingface.co/spacy/en_core_web_sm/resolve/main/en_core_web_sm-any-py3-none-any.whl"`

image

docs/hub/spacy.md
```diff
@@ -24,7 +24,7 @@ All models on the Hub come up with useful features
 All `spaCy` models from the Hub can be directly installed using pip install.
 
 ```bash
-pip install https://huggingface.co/spacy/en_core_web_sm/resolve/main/en_core_web_sm-any-py3-none-any.whl
+pip install "en_core_web_sm @ https://huggingface.co/spacy/en_core_web_sm/resolve/main/en_core_web_sm-any-py3-none-any.whl"
```
osanseviero, line 27: LGTM! You might want to update https://github.com/huggingface/huggingface.js/blob/main/packages/tasks/src/model-libraries-snippets.ts#L698 as well

osanseviero (approved):

PR #1465, opened 2024-10-22
Enterprise: document Analytics

gary149:
https://moon-ci-docs.huggingface.co/docs/hub/pr_1465/en/enterprise-hub-analytics
image


SBrandeis (approved):
Thank you

julien-c (commented):
late to the party, but looks great

maybe mention a note that we are considering adding unique users/IPs

PR #1445, opened 2024-10-14
Add playground to chat-completion doc

gary149:

PR #1468, opened 2024-10-23
Update chat-completion template doc to include API Playground section

hanouticelina:
Following up on PR #1445 which added the API Playground section to the generated `chat-completion.md` file, the `docs/api-inference/tasks/*.md` files should not be edited directly as they are automatically generated. 

this PR updates `chat-completion.handlebars` template to make those changes persistent by adding the "API Playground" section directly in the template. 
Note: All changes in other generated docs are from running the `generate` scripts.

cc @gary149 for viz.


Wauplin (approved):
Thanks for fixing this!

julien-c:
thx for fixing @hanouticelina!

PR #1471, opened 2024-10-25
fix: zsh no matches found error

p13rr0m:
ZSH treats square brackets as special characters and will respond with "no matches found: huggingface_hub[hf_transfer]".


julien-c (approved):

Vaibhavs10 (approved):
Thanks! 🤗

PR #1481, opened 2024-11-04
fix: image for Deploy To Spaces svg

mscolnick:
Updates the URLs to use `raw` instead of `resolve` which gets the svgs to show up properly.

### DIFF
--- a/docs/hub/spaces-sdks-docker-giskard.md
+++ b/docs/hub/spaces-sdks-docker-giskard.md
@@ -27,7 +27,7 @@ Hugging Face Spaces. This Giskard app is a **self-contained application complete
 You can deploy Giskard on Spaces with just a few clicks:
 
 <a  href="https://huggingface.co/new-space?template=giskardai%2Fgiskard">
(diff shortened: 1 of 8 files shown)


julien-c:
hmm this is weird, what is the issue with resolve? i think `resolve` is the intended way to do this but cc @coyotte508 for guidance

julien-c (commented):
ah yeah i see arf – we'll fix this upstream

image

coyotte508 (changes requested):
I don't think there's a problem for most users.

If you go on https://huggingface.co/docs/hub/spaces-sdks-docker-giskard in a private window it'll show fine.

The problem happens because you accessed the SVG directly, which "poisons" your cache with a wrong content-type - as we render different content-type depending on if served as an image or not (with the `sec-fetch-dest` header).

We could fix the problem at the root if we find how to avoid this cache issue - but anyway no problem for 99.9% of the users - unless they accessed https://huggingface.co/datasets/huggingface/badges/resolve/main/deploy-to-spaces-lg.svg directly first

mscolnick:
Ah thanks for the explanation, I did indeed go to that url

PR #1523, opened 2024-12-11
Fix: exclude package files from automated updates

hanouticelina:
Fixes shell script syntax in the workflow, following #1519.


Wauplin (approved):
oops, thanks!

PR #1532, opened 2024-12-17
Update dask docs

lhoestq:
- link to blog post for real world use case
- explain how dask parallelism works
- recommend squash history after writing to HF
- add a distributed data processing example
- explain predicate and projection pushdown + example

### DIFF
--- a/docs/hub/datasets-dask.md
+++ b/docs/hub/datasets-dask.md
@@ -1,7 +1,14 @@
 # Dask
 
 [Dask](https://github.com/dask/dask) is a parallel and distributed computing library that scales the existing Python and PyData ecosystem.
-Since it uses [fsspec](https://filesystem-spec.readthedocs.io) to read and write remote data, you can use the Hugging Face paths ([`hf://`](/docs/huggingface_hub/guides/hf_file_system#integrations)) to read and write data on the Hub:
+
+In particular, we can use Dask DataFrame to scale up pandas workflows. Dask DataFrame parallelizes pandas to handle large tabular data. It closely mirrors the pandas API, making it simple to transition from testing on a single dataset to processing the full dataset. Dask is particularly effective with Parquet, the default format on Hugging Face Datasets, as it supports rich data types, efficient columnar filtering, and compression.
+
+A good practical user-case for Dask is to run data processing or model inference on a dataset in a distributed manner. See for example the excellent blog post on [Scaling AI-Based Data Processing with Hugging Face + Dask](https://huggingface.co/blog/dask-scaling) by Coiled.
+
+# Read and Write
+
+Since Dask uses [fsspec](https://filesystem-spec.readthedocs.io) to read and write remote data, you can use the Hugging Face paths ([`hf://`](/docs/huggingface_hub/guides/hf_file_system#integrations)) to read and write data on the Hub;
 
 First you need to [Login with your Hugging Face account](/docs/huggingface_hub/quick-start#login), for example using:
 
@@ -17,7 +24,8 @@ from huggingface_hub import HfApi
 HfApi().create_repo(repo_id="username/my_dataset", repo_type="dataset")
 ```
 
-Finally, you can use [Hugging Face paths](/docs/huggingface_hub/guides/hf_file_system#integrations) in Dask:
+Finally, you can use [Hugging Face paths](/docs/huggingface_hub/guides/hf_file_system#integrations) in Dask.
+Dask DataFrame supports distributed writing to Parquet on Hugging Face, which uses commits to track dataset changes:
 
 ```python
 import dask.dataframe as dd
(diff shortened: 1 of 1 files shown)


docs/hub/datasets-dask.md
```diff
@@ -1,7 +1,14 @@
 # Dask
 
 [Dask](https://github.com/dask/dask) is a parallel and distributed computing library that scales the existing Python and PyData ecosystem.
-Since it uses [fsspec](https://filesystem-spec.readthedocs.io) to read and write remote data, you can use the Hugging Face paths ([`hf://`](/docs/huggingface_hub/guides/hf_file_system#integrations)) to read and write data on the Hub:
+
+In particular, we can use Dask DataFrame to scale up pandas workflows. Dask DataFrame parallelizes pandas to handle large tabular data. It closely mirrors the pandas API, making it simple to transition from testing on a single dataset to processing the full dataset. Dask is particularly effective with Parquet, the default format on Hugging Face Datasets, as it supports rich data types, efficient columnar filtering, and compression.
+
+A good practical use-case for Dask is to run data processing or model inference on a dataset in a distributed manner. See for example the excellent blog post on [Scaling AI-Based Data Processing with Hugging Face + Dask](https://huggingface.co/blog/dask-scaling) by Coiled.
```
davanstrien, line 7:
```suggestion
A good practical use case for Dask is running data processing or model inference on a dataset in a distributed manner. See, for example, Coiled's excellent blog post on [Scaling AI-Based Data Processing with Hugging Face + Dask](https://huggingface.co/blog/dask-scaling).
```

docs/hub/datasets-dask.md
```diff
@@ -45,3 +61,56 @@ df_test  = dd.read_parquet("hf://datasets/username/my_dataset/test")
...
+
+And in Dask you can run this function on every partition:
+
+```python
+# Dask API: run the function on every partition
+df["num_words"] = df.text.map_partitions(dummy_count_words, meta=int)
+```
+
+Note that you also need to provide `meta` which is the type of the pandas Series or DataFrame in the output of your function.
+This is needed because Dask DataFrame is a lazy API. Since Dask will only run the data processing once `.compute()` is called, it needs
```
davanstrien, line 89:
```suggestion
This is needed because Dask DataFrame uses a lazy API. Since Dask will only run the data processing once `.compute()` is called, it needs
```

docs/hub/datasets-dask.md
```diff
@@ -45,3 +61,56 @@ df_test  = dd.read_parquet("hf://datasets/username/my_dataset/test")
...
+
+For example this subset of FineWeb-Edu contains many Parquet files. If you can filter the dataset to keep the text from recent CC dumps, Dask will skip most of the files and only download the data that match the filter:
+
+```python
+import dask.dataframe as dd
+
+df = dd.read_parquet("hf://datasets/HuggingFaceFW/fineweb-edu/sample/10BT/*.parquet")
+
+# Dask will skip the files or row groups that don't
+# match rhe query without downloading them.
```
davanstrien, line 106:
```suggestion
# match the query without downloading them.
```

docs/hub/datasets-dask.md
```diff
@@ -45,3 +61,56 @@ df_test  = dd.read_parquet("hf://datasets/username/my_dataset/test")
...
+df = df[df.dump >= "CC-MAIN-2023"]
```
davanstrien, line 107: Does Dask not still need to download the data to check the values in this column match this filter? From what I understood in the Polars case the predicate push down is usually used for skipping the reading of a column i.e. if you drop it later it doesn't bother to load it and/or doing a filtering step early on. Is Dask directly able to do this before loading?

davanstrien (approved):
Looks great! If you don't have a chance I can also add something similar for Polars. I had one question about how Dask does filtering for the downloads. I probably misunderstood something here.

docs/hub/datasets-dask.md
lhoestq: so it will skip the row groups which don't have any row that match the query using the row group metadata

then on the remaining row groups it will download the column used for filtering to apply the filter

the other columns will be downloaded or not based on the other operations done on the dataset

lhoestq: > if you drop it later it doesn't bother to load it and/or doing a filtering step early on. Is Dask directly able to do this before loading?

yes correct ! would be cool to explain that here as well

docs/hub/datasets-dask.md
```diff
@@ -45,3 +61,56 @@ df_test  = dd.read_parquet("hf://datasets/username/my_dataset/test")
...
+
+Note that you also need to provide `meta` which is the type of the pandas Series or DataFrame in the output of your function.
+This is needed because Dask DataFrame uses a lazy API. Since Dask will only run the data processing once `.compute()` is called, it needs
+the `meta` argument to know the type of the new column in the meantime.
+
+# Predicate and Projection Pushdown
+
+When reading Parquet data from Hugging Face, Dask automatically leverages the metadata in Parquet files to skip entire files or row groups if they are not needed. For example if you apply a filter (predicate) on a Hugging Face Dataset in Parquet format or if you select a subset of the columns (projection), Dask will read the metadata of the Paquet files to discard the parts that are not needed without downloading them.
+
+This is possible thanks to the `dask-expr` package which is generally installed by default with Dask.
```
davanstrien, line 96:
```suggestion
This is possible thanks to the `dask-expr` package, which is generally installed by default with Dask. You can read more about `dask-expr` in this [blog post](https://blog.dask.org/2023/08/25/dask-expr-introduction). 
```

lhoestq:
thanks for the review ! merging this one for now but lmk if you have more comments

PR #1535, opened 2024-12-18
Hub datasets

julien-c:


docs/hub/datasets-downloading.md
```diff
@@ -2,7 +2,7 @@
 
 ## Integrated libraries
 
-If a dataset on the Hub is tied to a [supported library](./datasets-libraries), loading the dataset can be done in just a few lines. For information on accessing the dataset, you can click on the "Use in dataset library" button on the dataset page to see how to do so. For example, [`samsum`](https://huggingface.co/datasets/samsum?library=true) shows how to do so with 🤗 Datasets below.
+If a dataset on the Hub is tied to a [supported library](./datasets-libraries), loading the dataset can be done in just a few lines. For information on accessing the dataset, you can click on the "Use this dataset" button on the dataset page to see how to do so. For example, [`samsum`](https://huggingface.co/datasets/Samsung/samsum?library=datasets) shows how to do so with 🤗 Datasets below.
```
julien-c, line 5: hmm `Samsung/samsum` is not listed as compatible with any library right now for some reason (no "Use this dataset" button)

severo (approved):

docs/hub/datasets-downloading.md
severo: yes: it's a script-based dataset. We don't support them anymore.

julien-c: but even in the datasets library?

severo: hmmm good point

julien-c: ok, maybe not worth it then

maybe let's just switch to another dataset in the doc then

PR #1538, opened 2024-12-19
Add general example of split/subset on the main "Data Files Configuration" page

lhoestq:
This way users get an idea of what it means in practice without having to click on the documentation on automatic split detection or on the documentation on manual configuration of splits and subsets

I also moved the Tip on setting a default configuration in its right paragraph


julien-c (approved):
nice

Vaibhavs10 (approved):
🔥

PR #1541, opened 2024-12-20
docs: update inference API examples for reranker model

Muhtasham:
- Update code examples to show correct input format for reranker
- Add examples for text/text_pair format
- Include multiple input pairs to demonstrate batch processing
- Update curl, Python and JavaScript examples consistently

### DIFF
--- a/docs/api-inference/tasks/text-classification.md
+++ b/docs/api-inference/tasks/text-classification.md
@@ -37,6 +37,7 @@ Explore all available models and find the one that suits you best [here](https:/
 <inferencesnippet>
 
 <curl>
+  
 ```bash
 curl https://api-inference.huggingface.co/models/mixedbread-ai/mxbai-rerank-base-v1 \
     -X POST \
@@ -44,9 +45,11 @@ curl https://api-inference.huggingface.co/models/mixedbread-ai/mxbai-rerank-base
     -H 'Content-Type: application/json' \
     -H 'Authorization: Bearer hf_***'
 ```
+
 </curl>
 
 <python>
+  
 ```python
 import requests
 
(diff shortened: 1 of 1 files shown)


docs/api-inference/tasks/text-classification.md
```diff
@@ -38,63 +38,66 @@ Explore all available models and find the one that suits you best [here](https:/
...
 output = query({
-	"inputs": "I like you. I love you",
+    "inputs": [
+        {"text": "What is Paris?", "text_pair": "Paris is the capital of France"},
+        {"text": "What is Paris?", "text_pair": "London is the capital of England"}
+    ]
 })
 ```
-
-To use the Python client, see `huggingface_hub`'s [package reference](https://huggingface.co/docs/huggingface_hub/package_reference/inference_client#huggingface_hub.InferenceClient.text_classification).
```
Vaibhavs10, line 65: Could we use a more relevant package ref, like: https://huggingface.co/docs/huggingface_hub/package_reference/inference_client#huggingface_hub.InferenceClient.sentence_similarity

instead of removing it.

Vaibhavs10 (commented):
Thanks for the update! @tomaarsen would you like to give it a look too please! 🙏

docs/api-inference/tasks/text-classification.md
tomaarsen: Sentence Similarity works a bit differently behind the scenes sadly, I believe. Text Classification (of pairs) is actually what's happening behind the scenes.

tomaarsen (commented):
I'm not sure we should remove the mention of https://huggingface.co/docs/huggingface_hub/package_reference/inference_client#huggingface_hub.InferenceClient.text_classification, but otherwise I'm very much in favor of 1) the new model choice and 2) updated inputs.

docs/api-inference/tasks/text-classification.md
Vaibhavs10: Perfect! Thanks for the clarification. In that case, let's keep it.

Vaibhavs10: Hey @Muhtasham - can you update the PR to keep the ref to HFH lib?

Vaibhavs10 (approved):
Thanks for iterating here @Muhtasham 🤗

PR #1547, opened 2024-12-27
Update mlx-image.md

laclouis5:
This PR removes an outdated warning about grouped conv2d, which is now supported by MLX.


Vaibhavs10 (approved):
Thanks for the PR @laclouis5 🤗

PR #1548, opened 2024-12-30
docs: add Resource Groups API documentation

Kakulukian:
Public api for resource group


Pierrci (approved):
nice!

julien-c (commented):
cool!

I think we also wanted to add the API docs about orgs' user management, @Kakulukian @Pierrci ?

Pierrci:
ah yes, if it's not there already it would probably make sense to add it too, indeed!

julien-c:
@Kakulukian can you update this part here to link or copy/paste your API doc?

image

PR #1561, opened 2025-01-07
Spaces: Update 'Run with Docker' to 'Run locally'

kefranabg:
Reflect the updates from [this PR](https://github.com/huggingface-internal/moon-landing/pull/12145) in the hub documentation


docs/hub/_toctree.yml
```diff
@@ -291,8 +291,8 @@
         title: Langfuse on Spaces
   - local: spaces-embed
     title: Embed your Space
-  - local: spaces-run-with-docker
-    title: Run Spaces with Docker
+  - local: spaces-run-locally
+    title: Run locally
```
julien-c, line 295: hmm imo no need to change this (and it breaks an existing URL)

PR #1617, opened 2025-02-25
[docs] Model repo files

stevhliu:
Based on discussion [here](https://huggingface.slack.com/archives/C08DFQRED1A/p1740390537157329), this PR adds a doc explaining the files inside a model repo.


julien-c (changes requested):
hmm i'm not sure we want to be too focused on just transformers and diffusers

If anything this should be folded under each library in https://huggingface.co/docs/hub/models-libraries

stevhliu:
Good point! I've split up and added the content under each library doc instead

julien-c (commented):
ah sorry for the delay @stevhliu, this was fine i think, feel free to re-open and merge it

lgtm

julien-c (approved):

PR #1632, opened 2025-03-09
docs(hub-api): add expand param to get models

Johann-S:
Hi 👋 

I was using your API and I found `expand` parameter was missing in the API documentation but not on the Python lib (https://github.com/huggingface/huggingface_hub/blob/v0.29.2/src/huggingface_hub/hf_api.py#L1738)


julien-c (commented):
lgtm, i would also document it in other endpoints too (datasets and spaces)?

asking @coyotte508 for a review

PR #1640, opened 2025-03-17
set a `/docs/hub/xet` alias for ease of use

julien-c:
and also add a note that we will fast-track PRO users and Enterprise Hub organizations in the waitlist


jsulz (approved):
Nice - LGTM!

PR #1641, opened 2025-03-17
More multimodal datasets docs

lhoestq:
more docs for image/audio/webdataset

+ added docs for video

### DIFF
--- a/docs/hub/datasets-image.md
+++ b/docs/hub/datasets-image.md
@@ -218,4 +218,4 @@ dataset_info:
     dtype: string
 ```
 
-Note that Parquet is recommended for small images (<1MB per image) and small row groups (100 rows per row group, which is what `datasets` uses for images). For larger images it is recommended to use the WebDataset format, or to share the original image files (optionally with metadata files).
+Note that Parquet is recommended for small images (<1MB per image) and small row groups (100 rows per row group, which is what `datasets` uses for images). For larger images it is recommended to use the WebDataset format, or to share the original image files (optionally with metadata files, and following the [repositories recommendations and limits](https://huggingface.co/docs/hub/en/storage-limits) for storage and number of files).


docs/hub/datasets-data-files-configuration.md
```diff
@@ -48,12 +48,13 @@ See the documentation on [Manual configuration](./datasets-manual-configuration)
 
 See the [File formats](./datasets-adding#file-formats) doc page to find the list of supported formats and recommendations for your dataset. If your dataset uses CSV or TSV files, you can find more information in the [example datasets](https://huggingface.co/collections/datasets-examples/format-csv-and-tsv-655f681cb9673a4249cccb3d).
 
-## Image and Audio datasets
+## Image, Audio and Video datasets
 
-For image and audio classification datasets, you can also use directories to name the image and audio classes.
-And if your images/audio files have metadata (e.g. captions, bounding boxes, transcriptions, etc.), you can have metadata files next to them.
+For image/audio.video classification datasets, you can also use directories to name the image/audio/video classes.
```
davanstrien, line 53:
```suggestion
For image/audio/video classification datasets, you can also use directories to name the image/audio/video classes.
```

docs/hub/datasets-image.md
```diff
@@ -213,4 +218,4 @@ dataset_info:
     dtype: string
 ```
 
-Alternatively, Parquet files with Image data can be created using the `datasets` library by setting the column type to `Image()` and using the `.to_parquet(...)` method or `.push_to_hub(...)`. You can find a guide on loading image datasets in `datasets` [here](/docs/datasets/image_load).
+Note that Parquet is recommended for small images (<1MB per image) and small row groups (100 rows per row group, which is what `datasets` uses for images). For larger images it is recommended to use the WebDataset format, or to share the original image files (optionally with metadata files).
```
davanstrien, line 221: Just to clarify, does the guidance here say that parquet makes sense only if you have small images AND small row groups? I suspect more people will default to original image files so maybe we also need to point to the repository limits page too in that case?

docs/hub/datasets-webdataset.md
```diff
@@ -18,6 +18,21 @@ Labels and metadata can be in a `.json` file, in a `.txt` (for a caption, a desc
 A large scale WebDataset is made of many files called shards, where each shard is a TAR archive.
 Each shard is often ~1GB but the full dataset can be multiple terabytes!
 
+## Multimodal support
+
+WebDataset is designed for multimodal datasets, i.e. for image, audio and/or video datasets.
+
+Indeed since media files tend to be quite big, the sequential I/O of WebDataset enables large reads and buffering. This results in obtaining the best data loading speed.
```
davanstrien, line 25:
```suggestion
Indeed, since media files tend to be quite big, WebDataset's sequential I/O enables large reads and buffering, resulting in the best data loading speed.
```

docs/hub/datasets-webdataset.md
```diff
@@ -18,6 +18,21 @@ Labels and metadata can be in a `.json` file, in a `.txt` (for a caption, a desc
...
+
+Here is a non-exhaustive list of supported data formats:
+
+- image: jpeg, png, tiff
+- audio: mp3, m4a, wav, flac
+- video: mp4, mov, avi
+- other: npy, npz
+
+The full list evolves over time and depends on the implementation. For examoke you can can find which formats the `webdataset` package supports in the source code [here](https://github.com/webdataset/webdataset/blob/main/webdataset/autodecode.py).
```
davanstrien, line 34:
```suggestion
The full list evolves over time and depends on the implementation. For example, you can find which formats the `webdataset` package supports in the source code [here](https://github.com/webdataset/webdataset/blob/main/webdataset/autodecode.py).
```

davanstrien (approved):
Had one qs and small suggestions but looks good :)

julien-c (approved):

docs/hub/datasets-image.md
lhoestq: yup correct, I'll add a link to the repo limits

PR #1645, opened 2025-03-24
Update sagemaker docs structure

arjkesh:
Update sagemaker docs structure to remove references to old images, as well as structure in a way where automated PRs will update tables


docs/sagemaker/reference.md
arjkesh, line 1: this file is not showing up as a page anymore--I can include it as a section called "Examples" if helpful

pagezyhf: Yes I think we should keep this page for now. There's not only a list of notebook examples but also a useful doc on the inference toolkit.

pagezyhf: Not sure what makes sense for the title of the page. There are not only examples but also API specs, so I wouldn't call it "Examples". maybe the best is to keep "Reference" for now?

docs/sagemaker/tgi.md
```diff
@@ -0,0 +1,36 @@
...
+| Framework Version | Image Type | Image URI | Size (GB) | Pushed At | Details |
+| --- | --- | --- | --- | --- | --- |
+| 2.6 | gpu | `763104351884.dkr.ecr.us-west-2.amazonaws.com/huggingface-pytorch-tgi-inference:2.6.0-tgi3.1.1-gpu-py311-cu124-ubuntu22.04-v2.0` | 8.1 | 2025-03-17 16:47:39 | [Details](https://github.com/aws/deep-learning-containers/blob/master/available_images.md#huggingface-text-generation-inference-tgi-containers) |
+| 2.4 | gpu | `763104351884.dkr.ecr.us-west-2.amazonaws.com/huggingface-pytorch-tgi-inference:2.4.0-tgi3.0.1-gpu-py311-cu124-ubuntu22.04-v2.2` | 6.5 | 2025-03-06 18:28:24 | [Details](https://github.com/aws/deep-learning-containers/blob/master/available_images.md#huggingface-text-generation-inference-tgi-containers) |
+| 2.3 | gpu | `763104351884.dkr.ecr.us-west-2.amazonaws.com/huggingface-pytorch-tgi-inference:2.3.0-tgi2.2.0-gpu-py310-cu121-ubuntu22.04-v2.1` | 4.92 | 2024-10-04 21:59:12 | [Details](https://github.com/aws/deep-learning-containers/blob/master/available_images.md#huggingface-text-generation-inference-tgi-containers) |
+
+
+### SM Example
+```
+# create Hugging Face Model Class
```
pagezyhf, line 21: We should add the import in the code snippet

docs/sagemaker/transformers.md
```diff
@@ -0,0 +1,64 @@
...
+| Framework Version | Image Type | Image URI | Size (GB) | Pushed At | Details |
+| --- | --- | --- | --- | --- | --- |
+| 2.3 | gpu | `763104351884.dkr.ecr.us-west-2.amazonaws.com/huggingface-pytorch-training:2.3.0-transformers4.48.0-gpu-py311-cu121-ubuntu20.04-v2.1` | 8.75 | 2025-03-14 13:15:19 | [Details](https://github.com/aws/deep-learning-containers/blob/master/available_images.md#huggingface-training-containers) |
+
+
+### SM Example
+```
+# create Hugging Face Model Class
+huggingface_model = HuggingFaceModel(
+	image_uri=get_huggingface_llm_image_uri("huggingface",version="2.3"),
```
pagezyhf, line 21: The util doesn't work for Pytorch training  DLC, no?

docs/sagemaker/reference.md
arjkesh: Sure, we can keep it

arjkesh:
> Hello Arjuna, thanks for the PR! A few comments: I would keep all the containers info in a single page if possible, same as here: https://huggingface.co/docs/google-cloud/en/containers/available. This way, our customers don't need to know the name of our libraries to browse through the containers

Sounds reasonable. We can keep it in one file for now and then restructure later. That being said, if the headings in the same file are still "TGI", "Transformers", etc, customers will still need to know the library names to know what to use, similarly to if these are the titles of subheadings.

Is there a better way to denote the differences that might be more intuitive?

pagezyhf:
> Sounds reasonable. We can keep it in one file for now and then restructure later. That being said, if the headings in the same file are still "TGI", "Transformers", etc, customers will still need to know the library names to know what to use, similarly to if these are the titles of subheadings. Is there a better way to denote the differences that might be more intuitive?

Yes I think the list of containers deserve a short intro / how to section where you explain the rational on how to pick the right container for your usecase. 

For a basic intro, I'd say there are two main things to explain: 
- whether you want to do inference or training (Pytorch Training DLC vs others)
- knowing if your model is supported by an available inference toolkit DLC which should yield better perf than the Pytorch Inference DLC (generic purpose). A small decision tree could be very cool here. 

We could have a more complex version of the decision tree in an "advanced section". It could take into account more factors like quantization, target hardware etc.. This doc doesn't exist but becomes more necessary now TGI is multibackend.

pagezyhf:
Btw, I think the name we use aez Pytorch Inference DLC and Pytorch Training DLC over Transformers DLC. I don't know if it's a good name andwe can change it as long as we stay consistent.

PR #1654, opened 2025-03-27
Added min. version for huggingface_hub

rajatarya:
Adding min version of `huggingface_hub` to get `hf_xet` support.

### DIFF
--- a/docs/hub/storage-backends.md
+++ b/docs/hub/storage-backends.md
@@ -51,7 +51,7 @@ Note that our intent is to fast-track PRO users and Enterprise Hub organizations
 </Tip>
 
 
-To access a Xet-aware client, add the `hf_xet` Python package when installing `huggingface_hub`:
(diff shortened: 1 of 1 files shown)


jsulz (approved):
🚢  

LGTM!

docs/hub/storage-backends.md
```diff
@@ -57,7 +57,7 @@ To access a Xet-aware client, add the `hf_xet` Python package when installing `h
 pip install -U huggingface_hub[hf_xet]
 ```
 
-If you use the `transformers` or `datasets` libraries, it's already using `huggingface_hub` so you can simply install `hf_xet` in the same env:
+If you use the `transformers` or `datasets` libraries, it's already using `huggingface_hub` (`huggingface_hub` should be >= 0.30.0) so you can simply install `hf_xet` in the same env:
```
julien-c, line 60: maybe also add it at line 54 too?

julien-c (approved):
lgtm and cc'ing @Wauplin @hanouticelina

Wauplin (approved):

PR #1697, opened 2025-04-16
Add Cohere to the Providers table

hanouticelina:
follow-up PR to #1696, sorry forgot to mention that when reviewing the PR 😄


Wauplin (approved):
As yes, ofc

PR #1735, opened 2025-05-12
replace A100 with H200 in zero-gpu doc

linoytsaban:
small correction to update A100 -> H200


julien-c (approved):
good catch, was it the only mention of H100?

linoytsaban:
@julien-c yes the rest is H200 👍🏻

PR #1739, opened 2025-05-13
[Inference Providers] Fix generate + more instructions for providers documentation

hanouticelina:
This PR modifies how the list of inference providers is determined in the `generate` script.
Previously, `INFERENCE_PROVIDERS` was imported directly from `@huggingface/inference`. This list could potentially include staging providers, which can cause the generation script to fail. 

To ensure the script only processes "live" inference providers, `INFERENCE_PROVIDERS` is now derived from the keys of the `PROVIDERS_HUB_ORGS` object.

Most of the files in this PR are auto-generated by running the generate script. I also added more details in the instructions for adding provider documentation.


Wauplin (approved):
Thanks for fixing this!

PR #1762, opened 2025-06-03
Update datasets-spark.md

anuvrat:
Correcting the dataset name in the write() statement to be consistent with the documentation and the name in the image.


julien-c:
cc @lhoestq

PR #1767, opened 2025-06-05
Updating Xet usage docs

jsulz:
Updating docs now that Xet is the default for new users/orgs and `hf_xet` is a required dependency in `huggingface_hub`

Also adding a pointer to Xet environment variables in the recommendation section to make it easier to access.


julien-c (commented):
lgtm, nice!

cc @coyotte508 @Vaibhavs10 for quick reviews

pcuenca (approved):

PR #1791, opened 2025-06-22
fix: correct typos in DuckDB SQL docs

kilavvy:
## **Description:**
```markdown
## 🎯 Overview
This PR corrects minor typographical errors found in the `datasets-duckdb-sql.md` documentation file.

## ✨ Changes Made
- Corrected the spelling of "Confectionery"
```


julien-c:
This is a typo from the source data set, so I don't think it makes sense to fix it

burtenshaw:
Thanks for the contribution @kilavvy , but we won't merge this because the typo comes from a dataset (as mentioned).

PR #1802, opened 2025-07-01
restructure inference providers index page

burtenshaw:
This PR restructures the index page of inference providers to make it easier to get started.

### DIFF
--- a/docs/inference-providers/index.md
+++ b/docs/inference-providers/index.md
@@ -71,10 +71,12 @@ Inference Providers requires passing a user token in the request headers. You ca
 
 For more details about user tokens, check out [this guide](https://huggingface.co/docs/hub/en/security-tokens).
 
+### Quick Start
+
 <hfoptions id="inference-providers-examples">
 <hfoption id="curl">
 
-### cURL
+**cURL**
 
 Let's start with a cURL command highlighting the raw HTTP request. You can adapt this request to be run with the tool of your choice.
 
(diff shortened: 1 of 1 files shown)


docs/inference-providers/index.md
```diff
@@ -53,9 +32,29 @@ Inference Providers offers a fast and simple way to explore thousands of models
 
 To get started quickly with [Chat Completion models](http://huggingface.co/models?inference_provider=all&sort=trending&other=conversational), use the [Inference Playground](https://huggingface.co/playground) to easily test and compare models with your prompts.
 
-
 <a href="https://huggingface.co/playground" target="blank"><img src="https://cdn-uploads.huggingface.co/production/uploads/5f17f0a0925b9863e28ad517/9_Tgf0Tv65srhBirZQMTp.png" style="max-width: 550px; width: 100%;"/></a>
 
+## Partners
```
Vaibhavs10: It was better up at the start, more to the point no?

time to useful info should be least and if a person does not know what IP is for they can scroll down.

Vaibhavs10 (approved):

docs/inference-providers/index.md
burtenshaw: > time to useful info should be least and if a person does not know what IP is for they can scroll down.

yeah. this is what I'm trying to achieve.

I see the features section as more important than the partners table. But it's a matter of opinion I guess.

SBrandeis (commented):
Nice ! Seems to break the ToC on the right side though 
image

burtenshaw:
> Nice ! Seems to break the ToC on the right side though image

nice catch. I dropped the titles so the languages won't appear on the ToC.

PR #1823, opened 2025-07-08
more doc consistency everywhere

julien-c:
it's good to re-read the whole doc from time to time


julien-c:
self-approving this

SBrandeis (commented):
👍

PR #1827, opened 2025-07-09
Adding dark&light versions of first-api-call.md thumbnail

Reeray:


burtenshaw:
@Vaibhavs10 @SBrandeis Do we want to put this graphic on the index page?

billing

stevhliu:
I think it may be more appropriate for the Pricing and Billing doc since the graphic is focused subscription and billing.

burtenshaw:
> I think it may be more appropriate for the Pricing and Billing doc since the graphic is focused subscription and billing.

@stevhliu it's definitely about billing, but the idea was that this graphic makes the user lifecycle succinct.

stevhliu:
Ah ok, thanks for clarifying! I feel like theres still a bit of disconnect between the graphic and the doc. When I see that graphic, thats what I'm expecting to be doing on this doc but then it is about finding a model on the Hub, trying the widget, and then copying the code snippet.

PR #1829, opened 2025-07-09
Fix pandas_audio_methods code snippet

khdlr:
Fixed the example code snippet which was incorrectly importing `SFMethods` from `pandas_image_methods` rather than `pandas_audio_methods`.


lhoestq (approved):
good catch !

PR #1859, opened 2025-08-05
Extra clarity for storage limits

davanstrien:
A few users thought storage limits were only for datasets. Small additional tip to make repo limit info clearer for users.


Vaibhavs10 (approved):
ok for me, but lets wait for @Pierrci to say LGTM too

PR #1874, opened 2025-08-15
feat(docs): Add documentation for EU-specific gated models

Charlie-Boyer:
This PR adds a new section to the models-gated.md documentation, explaining the extra_gated_eu_disallowed feature.
Context [](https://huggingface.slack.com/archives/C02EMARJ65P/p1755258926052309)


Vaibhavs10 (approved):

docs/hub/models-gated.md
```diff
@@ -185,3 +185,21 @@ Alternatively, you can programmatically login using `login()` in a notebook or a
 You can also provide the `token` parameter to most loading methods in the libraries (`from_pretrained`, `hf_hub_download`, `load_dataset`, etc.), directly from your scripts.
 
 For more details about how to login, check out the [login guide](https://huggingface.co/docs/huggingface_hub/quick-start#login).
+
+### Restricting Access for EU Users
+
+For gated models, you can add an additional layer of access control to specifically restrict users from European Union countries. This is useful if your model's license or terms of use prohibit its distribution in the EU.
```
pcuenca, line 191:
```suggestion
Gated repositories can be configured with an additional layer of access control to disallow access from European Union countries. This is useful if your license or terms of use prohibit distribution in the EU.
```

Not specific to models, I think (even though the rest of file does mention models everywhere).

pcuenca (approved):
Thank you!

docs/hub/models-gated.md
Charlie-Boyer: this is the model doc, we have another one for datasets [https://huggingface.co/docs/hub/en/datasets-gated](https://huggingface.co/docs/hub/en/datasets-gated)
maybe I should add this info here also, wdyt @Vaibhavs10 ?

pcuenca: Ah yes, you're right! Sure, we can do that here or in a follow-up PR, as you wish.

Charlie-Boyer: I added it

julien-c:
(nit) not  a big fan of documenting this feature publicly

PR #1876, opened 2025-08-19
[inference providers] reorder menu 

burtenshaw:
This PR reorders the menu in inference providers.

(x-posting [internal slack convo](https://huggingface.slack.com/archives/C089DMDB415/p1755623611192109?thread_ts=1755623069.675539&cid=C089DMDB415))

### DIFF
--- a/docs/inference-providers/_toctree.yml
+++ b/docs/inference-providers/_toctree.yml
@@ -22,12 +22,8 @@
   - local: guides/gpt-oss
     title: How to use OpenAI gpt-oss
 
-- title: API Reference
+- title: Tasks
   sections:
-  - local: tasks/index
-    title: Index
-  - local: hub-api
-    title: Hub API
   - title: Popular Tasks
     sections:
     - local: tasks/chat-completion
@@ -39,7 +35,7 @@
     - local: tasks/text-to-video
       title: Text to Video
   - title: Other Tasks
-    isExpanded: true
+    isExpanded: False
     sections:
     - local: tasks/audio-classification
       title: Audio Classification
@@ -73,7 +69,7 @@
       title: Translation
     - local: tasks/zero-shot-classification
       title: Zero Shot Classification
-      
+          
 - title: Providers
   sections:
   - local: providers/cerebras
@@ -104,6 +100,13 @@
     title: SambaNova
   - local: providers/together
     title: Together
+
+- title: API Reference
+  sections:
+  - local: tasks/index
+    title: Index
+  - local: hub-api
+    title: Hub API
 
 - local: register-as-a-provider
   title: Register as an Inference Provider
--- a/docs/inference-providers/_toctree.yml
+++ b/docs/inference-providers/_toctree.yml
@@ -6,10 +6,21 @@
     title: Pricing and Billing
   - local: hub-integration
     title: Hub integration
-  - local: register-as-a-provider
-    title: Register as an Inference Provider
   - local: security
     title: Security
+
+- title: Guides
+  sections:
+  - local: guides/first-api-call
+    title: Your First API Call
+  - local: guides/building-first-app
+    title: Building Your First AI App
+  - local: guides/structured-output
+    title: Structured Outputs with LLMs
+  - local: guides/function-calling
+    title: Function Calling
+  - local: guides/gpt-oss
+    title: How to use OpenAI gpt-oss
 
 - title: Providers
   sections:
@@ -42,20 +53,6 @@
   - local: providers/together
     title: Together
 
-- title: Guides
-  sections:
-  - local: guides/first-api-call
-    title: Your First API Call
-  - local: guides/building-first-app
-    title: Building Your First AI App
-  - local: guides/structured-output
-    title: Structured Outputs with LLMs
-  - local: guides/function-calling
-    title: Function Calling
-  - local: guides/gpt-oss
-    title: How to use OpenAI gpt-oss
-
-
 - title: API Reference
   sections:
   - local: tasks/index
@@ -73,7 +70,7 @@
     - local: tasks/text-to-video
       title: Text to Video
   - title: Other Tasks
-    isExpanded: false
+    isExpanded: true
     sections:
     - local: tasks/audio-classification
       title: Audio Classification
@@ -107,3 +104,6 @@
       title: Translation
     - local: tasks/zero-shot-classification
       title: Zero Shot Classification
+
+- local: register-as-a-provider
+  title: Register as an Inference Provider
--- a/docs/inference-providers/_toctree.yml
+++ b/docs/inference-providers/_toctree.yml
@@ -22,37 +22,6 @@
   - local: guides/gpt-oss
     title: How to use OpenAI gpt-oss
 
-- title: Providers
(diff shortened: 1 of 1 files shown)


docs/inference-providers/_toctree.yml
```diff
@@ -6,11 +6,22 @@
...
+    title: Your First API Call
+  - local: guides/building-first-app
+    title: Building Your First AI App
+  - local: guides/structured-output
+    title: Structured Outputs with LLMs
+  - local: guides/function-calling
+    title: Function Calling
+  - local: guides/gpt-oss
+    title: How to use OpenAI gpt-oss
+
```
NielsRogge, line 24: Would also put the task guides (like "chat completion", "text-to-image") below the guides, rather than below the individual providers.

burtenshaw: So that would mean moving the API reference up? That's in here: 913c9d35419b029ba931081653c1f3362b86dbdd

NielsRogge: Well, technically I only consider the "index" and "hub API" to be API reference, those can indeed be the final section. However the "popular tasks" like "chat completion", "feature extraction" etc. aren't API reference to me (they explain the various tasks, so they belong to "guides" imo rather than "API reference")

burtenshaw: Yeah. That makes sense to me. I'll order it as you suggest and then hang on for another review, just because it's moved around a few times already. 

it's in 77b82305f42bb27cdf59e2fca26c4d839f635a98

NielsRogge (approved):
Thanks! Next up requests: add image-to-image and VLM task guides.

pcuenca (commented):
It's a bit weird to me that `API Reference | Index` just points to tasks that are now located several levels above. Perhaps we could do something like:

- Add `tasks/index` back to the first item of the `Tasks` section. Change the title from `API Reference` to something like `Inference Tasks`.
- Keep the `Hub API` item last, if required, but call the section `Hub API` instead.

Vaibhavs10 (commented):
thought about it a bit, I don't have strong intuition about the order, would let @julien-c @Wauplin @hanouticelina chime in.

(probably better to wait till at least Wednesday before merge)

tasal9:
This pull request reorganizes the `docs/inference-providers/_toctree.yml` file to improve the structure and navigation of the documentation. The main changes include moving the Providers section and related pages to the end of the file, grouping API reference and provider registration pages together, and ensuring consistent formatting for expanded sections.

Documentation structure reorganization:

* Moved the Providers section and all individual provider pages (`providers/cerebras`, `providers/cohere`, etc.) to the end of the file for better logical grouping. [[1]](diffhunk://#diff-e44c93d52a29f85b8dccbe7c3f19421b4be9bf2fb3b7aba57d4923fc7cb7f60fL9-L44) [[2]](diffhunk://#diff-e44c93d52a29f85b8dccbe7c3f19421b4be9bf2fb3b7aba57d4923fc7cb7f60fR71-R107)
* Relocated the Hub API and Register as an Inference Provider pages to appear after the Providers section. [[1]](diffhunk://#diff-e44c93d52a29f85b8dccbe7c3f19421b4be9bf2fb3b7aba57d4923fc7cb7f60fL9-L44) [[2]](diffhunk://#diff-e44c93d52a29f85b8dccbe7c3f19421b4be9bf2fb3b7aba57d4923fc7cb7f60fR71-R107)
* Merged the API Reference and Inference Tasks sections, renaming "Popular Tasks" to "Inference Tasks" and reorganizing their order for clarity.
* Corrected the formatting of the `isExpanded` field in the Other Tasks section to use a capitalized "False" for consistency.

docs/inference-providers/_toctree.yml
```diff
@@ -106,4 +67,41 @@
...
+  - local: providers/sambanova
+    title: SambaNova
+  - local: providers/together
+    title: Together
+
+- local: hub-api
+  title: Hub API
+
+- local: register-as-a-provider
+  title: Register as an Inference Provider
```
Wauplin, line 107:
```suggestion
- title: Advanced
  sections:
    - local: hub-api
      title: Hub API
    - local: register-as-a-provider
      title: Register as an Inference Provider
```

I would put these two in a grouped section for consistency. I'm fine with something like "Other" or "Misc" as well

docs/inference-providers/_toctree.yml
```diff
@@ -55,25 +22,19 @@
   - local: guides/gpt-oss
     title: How to use OpenAI gpt-oss
 
-
-- title: API Reference
+- local: tasks/index
+  title: Inference Tasks
```
Wauplin, line 26: Makes perfect sense to rename this section, thanks for the suggestion :+1:

docs/inference-providers/_toctree.yml
```diff
@@ -106,4 +67,41 @@
     - local: tasks/translation
       title: Translation
     - local: tasks/zero-shot-classification
-      title: Zero Shot Classification
\ No newline at end of file
+      title: Zero Shot Classification
+          
+- title: Providers
```
Wauplin, line 72: no strong opinion whether to have "providers" below or above "tasks". Previous logic was to display our partners prominently but fine with moving them down if you prefer

Wauplin (approved):
Thanks @burtenshaw for opening this and @NielsRogge @pcuenca @Vaibhavs10 for the suggestions. Current proposition looks good to me expect the last two items that I would group in a new section. No strong opinion on the rest so pre-approving right now

docs/inference-providers/_toctree.yml
julien-c: agree

docs/inference-providers/_toctree.yml
julien-c: what about keeping "API Reference"? otherwise Advanced sounds fine to me

docs/inference-providers/_toctree.yml
julien-c: fine for me as well as long as Providers are not too hidden. Given Tasks is short, i'm fine with it being above

julien-c (approved):

docs/inference-providers/_toctree.yml
Vaibhavs10: API Reference sounds tad bit nicer to me.

docs/inference-providers/_toctree.yml
```diff
@@ -55,25 +22,19 @@
...
-    - local: tasks/text-to-image
-      title: Text to Image
-    - local: tasks/text-to-video
-      title: Text to Video
+  - local: tasks/chat-completion
+    title: Chat Completion
+  - local: tasks/feature-extraction
+    title: Feature Extraction
+  - local: tasks/text-to-image
+    title: Text to Image
```
NielsRogge, line 33: I would also make Image-to-Image and VLMs separate task pages here (as follow-up PRs)

docs/inference-providers/_toctree.yml
julien-c: feels a bit weird to have them free-standing: 
image

PR #1883, opened 2025-08-27
removing EU storage region restriction

jsulz:
Removing this restriction from the documentation since it's no longer valid 🎉


Vaibhavs10 (approved):
Nice!

julien-c:
nice!

PR #1952, opened 2025-09-24
xet protocol specification in hub docs

assafvayner:
Porting xet specification to final location in hub docs (from xet-core github).

Includes complete specification for xet protocol.

PR #1963, opened 2025-09-30
xet protocol files

assafvayner:
Adds xet-protocol files through hub-docs rather than xet-core. served out of the `docs/xet-protocol` path.

replaces https://github.com/huggingface/hub-docs/pull/1952

cc @rajatarya @jsulz


julien-c:
SGTM but maybe let's rename to /docs/xet, no? (given there might very well be more than just the spec in the future IMO)

cc @mishig25 for viz too

PR #1987, opened 2025-10-10
edit section of security doc for accuracy

Muhammadsaeed707:
from Luc Georges, "We don’t do scan triggers on page load anymore"


julien-c:
i thought the idea was to delegate ClamAV scanning to a third party partner, like ProtectAI and JFrog?

Pierrci:
I'm not sure this is the place to discuss this @julien-c, it's _slightly_ orthogonal

julien-c:
well if they scan, then they define when they scan. But ok to merge this in the meantime

PR #1969, opened 2025-10-02
Organize Hub Xet docs

rajatarya:
(...same changes as #1956 just put in new PR since messed up rebase...)

- reorganized Storage Backends page to be more Xet centric, have subpages, link to Xet Protocol pages.
- bumped Xet Protocol page to version 1.0.0
- Updated links to /storage-backends

PR #2010, opened 2025-10-23
Adding Git Xet docs

jsulz:
This PR brings docs about Git Xet into the core Hub docs. A few notes:
- I linked directly to Windows installers. Navigating GitHub to get to these downloads is clunky IMO. However, this creates link-update overhead on each release. Very open to alternatives if anyone has them.
- We're using a `homebrew tap` for Homebrew installs. We still need a PR to [homebrew-core](https://github.com/Homebrew/homebrew-core) to get a more simple `brew install`. @seanses plans to handle this in an upcoming release.
- Install-only guidance for now as deeper “how it works” may change soon.
- Only one pointer in `repositories-getting-started` to `git-xet` while we roll out. My feeling was to add more once it stabilizes, but also open to feedback here if we think it's better to add callouts more liberally. 

Also did some Markdown header cleanup from the restructure in https://github.com/huggingface/hub-docs/pull/1969 sorry for the extra changes cluttering up this PR as a result. 

Happy for any other feedback! 🤗

### DIFF
--- a/docs/hub/xet/using-xet-storage.md
+++ b/docs/hub/xet/using-xet-storage.md
@@ -52,11 +52,11 @@ Install [Git](https://git-scm.com/) and [Git LFS](https://git-lfs.com/).
    ```
 ### Windows (amd64)
 
- Using an installer: 
+Using an installer: 
  - Download `git-xet-windows-installer-x86_64.zip` ([available here](https://github.com/huggingface/xet-core/releases/download/git-xet-v0.1.0/git-xet-windows-installer-x86_64.zip)) and unzip. 
  - Run the `msi` installer file and follow the prompts.
- 
- Manual installation:
+
+Manual installation:
  - Download `git-xet-windows-x86_64.zip` ([available here](https://github.com/huggingface/xet-core/releases/download/git-xet-v0.1.0/git-xet-windows-x86_64.zip)) and unzip. 
  - Place the extracted `git-xet.exe` under a `PATH` directory.
  - Run `git-xet install` in a terminal.


docs/hub/repositories-getting-started.md
```diff
@@ -8,6 +8,9 @@ This document shows how to handle repositories through the web interface as well
 
 If you do not have `git` available as a CLI command yet, you will need to [install Git](https://git-scm.com/downloads) for your platform. You will also need to [install Git LFS](https://git-lfs.github.com/), which will be used to handle large files such as images and model weights.
 
+> [!TIP]
+> For improved upload and download speeds when working with large files and Git, install the [Git Xet](xet/using-xet-storage#git) extension.
```
julien-c, line 12: do we still need to install git-lfs for now? I think longer-term we should remove this need

julien-c: and BTW unrelated but is the Youtube video below still up-to-date? I haven't watched it in a while, could be useful to check

docs/hub/xet/index.md
julien-c, line 1: we should add a visual element to this page (maybe the speed GIF? or something else)

docs/hub/xet/using-xet-storage.md
```diff
@@ -24,19 +26,85 @@ To see more detailed usage docs, refer to the `huggingface_hub` docs for:
...
+
+Install [Git](https://git-scm.com/) and [Git LFS](https://git-lfs.com/). 
+
+### Install on macOS or Linux (amd64 or aarch64)
+
+ To install using [Homebrew](https://brew.sh/):
+   ```
+   brew tap huggingface/tap
+   brew install git-xet
+   ```
```
julien-c: mention we are working to get it into the default main tap?

docs/hub/xet/using-xet-storage.md
```diff
@@ -24,19 +26,85 @@ To see more detailed usage docs, refer to the `huggingface_hub` docs for:
...
 - [Managing the `hf_xet` cache](https://huggingface.co/docs/huggingface_hub/guides/manage-cache#chunk-based-caching-xet)
 
-### Recommendations
+## Git
+
+Git users can access the benefits of Xet by downloading and installing the Git Xet extension. Once installed, simply use the [standard workflows for managing Hub repositories with Git](../repositories-getting-started) - no additional changes necessary. 
+
+### Prerequisites
+
+Install [Git](https://git-scm.com/) and [Git LFS](https://git-lfs.com/).
```
julien-c, line 35: longer-term we'll remove the need for git-lfs correct?

julien-c (approved):

docs/hub/xet/using-xet-storage.md
```diff
@@ -24,19 +26,85 @@ To see more detailed usage docs, refer to the `huggingface_hub` docs for:
...
+### Install on macOS or Linux (amd64 or aarch64)
+
+ To install using [Homebrew](https://brew.sh/):
+   ```
+   brew tap huggingface/tap
+   brew install git-xet
+   ```
+ Or, using an installation script, run the following in your terminal (requires `curl` and `unzip`):
+   ```
+   curl --proto '=https' --tlsv1.2 -sSf https://raw.githubusercontent.com/huggingface/xet-core/refs/heads/main/git_xet/install.sh | sh
```
Vaibhavs10: I'd swap the order here: have the installation script first and the tap later.

docs/hub/xet/using-xet-storage.md
```diff
@@ -24,19 +26,85 @@ To see more detailed usage docs, refer to the `huggingface_hub` docs for:
...
-### Recommendations
+## Git
+
+Git users can access the benefits of Xet by downloading and installing the Git Xet extension. Once installed, simply use the [standard workflows for managing Hub repositories with Git](../repositories-getting-started) - no additional changes necessary. 
+
+### Prerequisites
+
+Install [Git](https://git-scm.com/) and [Git LFS](https://git-lfs.com/). 
+
+### Install on macOS or Linux (amd64 or aarch64)
```
Vaibhavs10, line 37: should we also somewhere guide people to as to how they can use it, git push/ pull? (just a small tutorial)

Vaibhavs10: could also be in this doc - 3 LoC

docs/hub/repositories-getting-started.md
jsulz: Currently `git-xet` is registered as a custom transfer agent on top of Git LFS, so it's still necessary to install Git LFS. 

Prior to the HF infra offsite, we had conversations about how to undo that dependency and need to revisit them, but long-term the goal is to indeed remove the need - cc @rajatarya

jsulz: I watched the video - it's still largely in line with the experience today and is nice to have IMO, but could probably use a refresh.

docs/hub/xet/index.md
jsulz: I believe @rajatarya was planning on moving some high-level diagrams here soon.

docs/hub/xet/using-xet-storage.md
jsulz: That's still the strategy (see my longer comment here https://github.com/huggingface/hub-docs/pull/2010#discussion_r2456091574)

docs/hub/xet/using-xet-storage.md
jsulz: Makes sense to me. Will add it.

docs/hub/xet/using-xet-storage.md
```diff
@@ -24,19 +26,103 @@ To see more detailed usage docs, refer to the `huggingface_hub` docs for:
...
+   ```
+
+  To verify the installation, run:
+   ```
+   git-xet --version
+   ```
+### Windows (amd64)
+
+ Using an installer: 
+ - Download `git-xet-windows-installer-x86_64.zip` ([available here](https://github.com/huggingface/xet-core/releases/download/git-xet-v0.1.0/git-xet-windows-installer-x86_64.zip)) and unzip.
```
rajatarya, line 56: This can be adjusted to have /latest/ in the URL - which should reduce the maintenance overhead. Unfortunately, GitHub only allows one latest release per repo, and we are planning on having two releases come from the same repo (hf-xet, git-xet). Not sure if we can address in a meaningful way.

rajatarya: One workaround could be that git-xet release process includes updating the tag for latest, so https://github.com/huggingface/xet-core/releases/tag/git-xet-latest would always point to the latest release tag.

That way docs don't need updating and the release process for git-xet manages where latest points.

docs/hub/xet/using-xet-storage.md
```diff
@@ -24,19 +26,103 @@ To see more detailed usage docs, refer to the `huggingface_hub` docs for:
...
 - **Prioritize community access**: Xet substantially increases the efficiency and scale of large file transfers. Instead of structuring your repository to reduce its total size (or the size of individual files), organize it for collaborators and community users so they may easily navigate and retrieve the content they need.
 
-### Current Limitations
+## Current Limitations
 
 While Xet brings fine-grained deduplication and enhanced performance to Git-based storage, some features and platform compatibilities are still in development. As a result, keep the following constraints in mind when working with a Xet-enabled repository:
 
-- **64-bit systems only**: The `hf_xet` client currently requires a 64-bit architecture; 32-bit systems are not supported.
-- **Git client integration (git-xet)**: Under active development - coming soon, stay tuned!
+- **64-bit systems only**: The `hf_xet` client currently requires a 64-bit architecture; 32-bit systems are not supported.
```
rajatarya: Both `hf_xet` and `git_xet` are for 64-bit platforms only.

docs/hub/xet/using-xet-storage.md
```diff
@@ -65,6 +66,23 @@ To verity the installation, run:
...
+
+Once installed on your platform, using Git Xet is as simple as following the Hub's standard Git workflows. First, make sure Git LFS is initialized in your local version of the repository: 
+
+  ```
+  git lfs install
+  ```
+Then make any additions you might want, commit your changes, and `push` your commit to the Hub:
+
+  ```
+  # Create any files you like! Then...
```
seanses, line 76: I think we should mention "git lfs track ..." here so not to confuse users and make them think that we aren't using the git lfs workflow.

jsulz: I would prefer to point to https://huggingface.co/docs/hub/repositories-getting-started#set-up for deeper instructions (which includes `git lfs track...`) rather than duplicate too much.

Does that sound reasonable?

docs/hub/xet/using-xet-storage.md
jsulz: If we're going to do anything, I would prefer use the tagging approach instead of /latest/ due to exactly what you are saying above. 

In the short term, I'll release the docs like this, but on our next release when a `git-xet`-specific tag is available, I'll update it.

docs/hub/xet/using-xet-storage.md
seanses: Yeah! Let's point to that doc

docs/hub/xet/using-xet-storage.md
jsulz: Nice catch! Got it.

docs/hub/xet/using-xet-storage.md
jsulz: Got it!

docs/hub/repositories-getting-started.md
julien-c: BTW maybe (small UX improvement) for now maybe `brew install git-xet` also installs `brew install git-lfs` as a dependency?

docs/hub/xet/using-xet-storage.md
julien-c: we could also do a basic lambda function somewhere ie. add an indirection layer (ie. resolve the URL outside of the doc) 

i don't know if you see what I mean lol

PR #2016, opened 2025-10-24
docs(hub-api): Add organization overview and followers endpoints to API documentation

BastienGimbert:
Hello,

Add two organization endpoints to the API docs.
`GET /api/organizations/{organization_name}/overview` (equivalent to `huggingface_hub.get_organization_overview()`) returns the org overview, 

and 

`GET /api/organizations/{organization_name}/followers` returns the org followers list.

I did not include the new “list_organization_followers” method that I developed because it is not yet available.   (https://github.com/huggingface/huggingface_hub/pull/3467)

EDIT : added list_organization_followers

### DIFF
--- a/docs/hub/api.md
+++ b/docs/hub/api.md
@@ -273,7 +273,7 @@ This is equivalent to `huggingface_hub.whoami()`.
 
 ## Organization API
 
-The following endpoint gets a list of the Organization members.
+The following endpoints handle organization like getting overview of an organization, listing members and followers.
 
 ### GET /api/organizations/{organization_name}/overview
 


docs/hub/api.md
```diff
@@ -275,6 +275,19 @@ This is equivalent to `huggingface_hub.whoami()`.
 
 The following endpoint gets a list of the Organization members.
```
julien-c: this should be kept near the relevant API call

BastienGimbert: That was the old title of the section, so I just changed it. 
New one :  "The following endpoints handle organization like getting overview of an organization, listing members and followers."

BastienGimbert:
> why not, but to be fair we plan on probably just removing this page in favor of just maintaining https://huggingface.co/spaces/huggingface/openapi (and adding the missing methods there)

I couldn't find the repo to modify the OpenAPI specification. I think it's not public.

julien-c:
> I couldn't find the repo to modify the OpenAPI specification. I think it's not public.


yes it's auto-generated from our codebase directly

julien-c (approved):

PR #2017, opened 2025-10-24
Update git-xet brew installation instruction

seanses:
Users need to run `git-xet install` after `brew install git-xet`. This instruction is printed out at the end of the `brew install git-xet` step but people may not pay attention to that.


pcuenca (approved):
Thank you!

julien-c (approved):

julien-c:
added some more here #2048 
(power of 2 PR number!! 🎉 )

PR #2026, opened 2025-10-30
[inference providers] add some more info in the "register as provider" docs

SBrandeis:


docs/inference-providers/register-as-a-provider.md
```diff
@@ -393,6 +396,39 @@ Inference-Id: unique-id-00131
 [response body]
 ```
 
+### Exposing pricing through OpenAI /models routes
+
+If your API is OpenAI-compatible, we kindly ask that you expose LLM pricing information and context length through the [`/v1/models` endpoint](https://platform.openai.com/docs/api-reference/models/list).
```
Vaibhavs10:
```suggestion
If your API is OpenAI-compatible, we expect that you expose LLM pricing information and context length through the [`/v1/models` endpoint](https://platform.openai.com/docs/api-reference/models/list).
```

docs/inference-providers/register-as-a-provider.md
```diff
@@ -393,6 +396,39 @@ Inference-Id: unique-id-00131
...
+
+This helps power our [provider comparison table](https://huggingface.co/inference/models) and other provider selection features.
```
Vaibhavs10:
```suggestion
This powers our [provider comparison table](https://huggingface.co/inference/models) and other provider selection features like `:cheapest` (which selects the cheapest provider for a model).
```

Vaibhavs10 (approved):
ty!

docs/inference-providers/register-as-a-provider.md
SBrandeis: 😈

julien-c:
late to the party but lgtm

PR #1765, opened 2025-06-04
update sso doc

Charlie-Boyer:

PR #2029, opened 2025-11-03
Fix unclosed HTML link tag in security-sso.md

mishig25:
## Fix unclosed HTML link tag in security-sso.md

Fixed an unclosed `<a>` tag in a WARNING admonition that was breaking markdown link rendering on the page.

**Changes:**
- Added missing `</a>` closing tag in `security-sso.md` line 8
- The unclosed HTML link within the blockquote was causing subsequent markdown links to not render properly

**File affected:**
- hub-docs/docs/hub/security-sso.md`

image


mishig25:
looking at past commits, the unclosed `</a>` tag was introduced in https://github.com/huggingface/hub-docs/pull/1765

image


ideally, we should use markdown links as well instead of html ones

PR #2047, opened 2025-11-12
Data studio a few renames

julien-c:


lhoestq:
Maybe we can rename "Dataset Viewer" -> "Data Studio" in index.md as well

PR #2056, opened 2025-11-14
just a few Large Files mentions

julien-c:
only where it really makes sense imo (it's fine to still mention LFS pointers etc for instance)

### DIFF
--- a/docs/hub/storage-limits.md
+++ b/docs/hub/storage-limits.md
@@ -73,7 +73,7 @@ There are a few reasons for this:
       happen when streaming data and smaller files avoid resuming from the beginning in case of errors.
     - Files are served to the users using CloudFront. From our experience, huge files are not cached by this service
       leading to a slower download speed.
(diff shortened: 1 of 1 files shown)


docs/hub/storage-limits.md
```diff
@@ -73,14 +73,14 @@ There are a few reasons for this:
       happen when streaming data and smaller files avoid resuming from the beginning in case of errors.
     - Files are served to the users using CloudFront. From our experience, huge files are not cached by this service
       leading to a slower download speed.
-In all cases no single LFS file will be able to be >50GB. I.e. 50GB is the hard limit for single file size.
+In all cases no single file will be able to be >500GB. I.e. 500GB is the hard limit for single file size.
```
davanstrien:
```suggestion
In all cases, no single file will exceed 500GB. I.e. 500GB is the hard limit for a single file size.
```

julien-c:
(note that there might remain other stuff to update in other pages, i haven't checked super exhaustively)

PR #2075, opened 2025-11-24
Update Spaces Billing

Michellehbn:
Quick update to text regarding billing

### DIFF
--- a/docs/hub/spaces-gpus.md
+++ b/docs/hub/spaces-gpus.md
@@ -119,7 +119,7 @@ print(tf.config.list_physical_devices('GPU'))
 Billing on Spaces is based on hardware usage and is computed by the minute: you get charged for every minute the Space runs on the requested hardware, 
 regardless of whether the Space is used.
 
-During a Space's lifecycle, it is only billed when the Space is actually `Starting` and `Running`. This means that there is no cost during build.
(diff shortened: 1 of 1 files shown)


docs/hub/spaces-gpus.md
```diff
@@ -119,7 +119,7 @@ print(tf.config.list_physical_devices('GPU'))
 Billing on Spaces is based on hardware usage and is computed by the minute: you get charged for every minute the Space runs on the requested hardware, 
 regardless of whether the Space is used.
 
-During a Space's lifecycle, it is only billed when the Space is actually `Running`. This means that there is no cost during build or startup.
+During a Space's lifecycle, it is only billed when the Space is actually `Starting` and `Running`. This means that there is no cost during build.
```
julien-c:
```suggestion
During a Space's lifecycle, it is only billed when the Space is `Starting` or `Running`. This means that there is no cost during build.
```

christophe-rannou:
```suggestion
Billing for a Space begins when it enters the Starting state and continues through the Running state. There is no cost during the build phase.
```

Michellehbn: Thanks, Chris! As discussed, we'll continue to update text if there is confusion from this change, if needed

PR #2082, opened 2025-11-26
Change 'Documentation' to 'Resources' in integrations table

davanstrien:
small clarity fix: Updated the column name reference from 'Documentation' to 'Resources'


pcuenca (approved):
Fine for me; the alternative would be to rename the column to "Documentation" as most of the items seem to be docs.

davanstrien:
> Fine for me; the alternative would be to rename the column to "Documentation" as most of the items seem to be docs.

We decided on "Resources" in another PR since we might point to guides etc too but we can change later if we decide it's nicer to use "Documentation" :)

PR #2093, opened 2025-12-04
[Inference providers] fix 404 links

mishig25:
[Inference providers] fix 404 links


hanouticelina:
I opened a PR to fix this in the generation script instead : #2094

PR #2094, opened 2025-12-04
[Inference Providers] Fix task links 

hanouticelina:
Follow-up to #2093. 
This PR fixes the task links in the inference providers docs. Note that almost all inference providers docs are auto-generated. Fixes should be made to the script or templates in `scripts/inference-providers/`, not to the generated markdown files directly.


julien-c:
thx!

PR #2106, opened 2025-12-11
Resource group wording

Muhammadsaeed707:
a resource group admin that is not an org admin can manage already existing repos in a resource group and administer the users of the resource group but they cannot add a repo to the resource group

reference: https://github.com/huggingface-internal/moon-landing/commit/2b66cac6226fac1a8287f63b4fe424c38dd8cd8a


coyotte508 (commented):
maybe add a wording that an admin of the *org* can transfer repos in/out?

Muhammadsaeed707:
> maybe add a wording that an admin of the _org_ can transfer repos in/out?

yes added ty

coyotte508 (approved):

PR #2107, opened 2025-12-11
Metadata proposal for `eval_results`

julien-c:
[internal link for context](https://github.com/huggingface-internal/moon-landing/issues/16109)

PR #2109, opened 2025-12-12
[enhancement] eval-results - add a page and menu

burtenshaw:
This PR adds the docs page for eval results, which should be useful for sharing with eval partners.

### DIFF
--- a/docs/hub/eval-results.md
+++ b/docs/hub/eval-results.md
@@ -73,7 +73,7 @@ Results display badges based on their metadata in the YAML file:
 | leaderboard | Links to the benchmark dataset |
 | source | Links to evaluation logs or external source |
 
-For more details on how to format this data, check out the [Eval Results](https://github.com/huggingface/huggingface_hub/blob/main/src/huggingface_hub/templates/eval_results.yaml) specifications.
+For more details on how to format this data, check out the [Eval Results](https://github.com/huggingface/hub-docs/blob/47a2b70cada8057a4acb02d8f19527ff16585ed1/eval_results.yaml) specifications.
 
 ### Community Contributions
 


docs/hub/_toctree.yml
```diff
@@ -179,6 +179,8 @@
     title: Model Inference
   - local: models-download-stats
     title: Models Download Stats
+  - local: eval-results
+    title: Evaluation Results
```
julien-c:
```suggestion
    title: Eval Results
```

maybe cleaner (we could change in the product actually @gary149)

julien-c: and maybe put it right after "Model cards"

docs/hub/eval-results.md
```diff
@@ -0,0 +1,83 @@
...
+
+![Benchmark Dataset](https://huggingface.co/huggingface/documentation-images/resolve/main/evaluation-results/benchmark-preview.png)
+
+### Registering a Benchmark
+
+To register your dataset as a benchmark:
+
+1. Create a dataset repo containing your evaluation data
+2. Add an `eval.yaml` file to the repo root with your benchmark configuration
+3. The file is validated at push time
```
julien-c, line 20: 4. there's currently an allow-list of datasets (while in Beta), so get in touch so we can add it

docs/hub/eval-results.md
```diff
@@ -0,0 +1,83 @@
...
+
+Or, with only the required attributes:
+
+```yaml
+- dataset:
+    id: Idavidrein/gpqa
+    config: gpqa_diamond
+  value: 0.412
+```
+
```
julien-c, line 66: maybe link to this file (ie. to where it will live) https://github.com/huggingface/hub-docs/pull/2107/changes as some kind of living spec document (like we do in https://huggingface.co/docs/hub/en/model-cards#evaluation-results for modelcard.md metadata)

julien-c (approved):

PR #2111, opened 2025-12-12
[Markdown] Document image theme context display

kefranabg:
Documenting how to display an image in Markdown depending on the theme

### DIFF
--- a/docs/hub/model-cards.md
+++ b/docs/hub/model-cards.md
@@ -297,7 +297,7 @@ Image only displays when viewing in dark mode
 
 **For already hosted images**
 
-If you want to reference images that are already hosted without re-uploading them, use HTML `<img>` tags with the following Tailwind CSS classes:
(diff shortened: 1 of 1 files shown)


kefranabg:
Ideally I would have preferred to display this in a page that is more dedicated to Markdown (since it works everywhere, not just in model card), but we don't have any page related to markdown syntax on the hub, so I guess the model card page is the best place right now for this content

julien-c:
yes i think it is a proper location to place it.

Another option is to create a new "Markdown formatting" page under Repositories (and put the LaTeX section + any other section that's common to all markdown on the Hub)

kefranabg:
> Another option is to create a new "Markdown formatting" page under Repositories (and put the LaTeX section + any other section that's common to all markdown on the Hub)

Yes I think that could be a good thing. Adding this to my TODO

PR #2159, opened 2026-01-15
add sso join link description

Muhammadsaeed707:
add description in docs for the new sso join link


Muhammadsaeed707:
> probably could be in https://huggingface.co/docs/hub/security-sso as well?

https://github.com/huggingface/hub-docs/pull/2167

PR #2167, opened 2026-01-19
add description to sso page as well

Muhammadsaeed707:
add description to sso docs as well for the new sso join link


Pierrci (approved):
lgtm 👍

PR #2178, opened 2026-01-22
[ZeroGPU] `xlarge` size

cbensimon:


docs/hub/spaces-zerogpu.md
```diff
@@ -87,6 +94,23 @@ gr.Interface(
...
+
+``` python
+@spaces.GPU(size="xlarge")
+def generate(prompt):
+    return pipe(prompt).images
+```
+
+> [!NOTE]
+> - `xlarge` consumes **2×** more daily quota than `large` (e.g. a 45s **effective** task duration consumes 90s of quota)
+> - `xlarge` usually means increased queuing probability and duration
```
pcuenca: Does this mean that xlarge may be assigned faster? I don't understand the comment about increased duration, I thought it was a hardcoded constant, or a value specified by the user; how are these affected by the type of slice?

cbensimon: Ok so we should probably reformulate.

> Does this mean that xlarge may be assigned faster?

No, it's the opposite. When specifying `xlarge` in `@spaces.GPU(size='xlarge')`:
1. There's more risk to get a "Waiting for a GPU to become available" (aka to be queued instead of instantly getting the GPU)
2. You always wait longer in the queue for an xlarge size compared to what you would have waited in the same conditions with a normal (aka `large`) size

> I don't understand the comment about increased duration, I thought it was a hardcoded constant, or a value specified by the user

Ok I see. Here with "duration" I was talking about queue wait time, not GPU task duration (indeed, specified by user or defaulting to 60s hard constant). I'll change the wording

cbensimon: Done in https://github.com/huggingface/hub-docs/pull/2178/changes/97d336a5273f4ff652b60e77e5ec97f534efd14f. Let me know what you think

PR #2187, opened 2026-01-27
[Jobs] Document labels

lhoestq:
requires https://github.com/huggingface/huggingface_hub/pull/3742 for labels filtering


lhoestq:
there are already docs on Jobs in the CLI doc page of huggingface_hub fwiw

PR #2200, opened 2026-02-03
Agents-section-tweaks

burtenshaw:
This pr makes two changes in the agents section:

- removes the old ;agents on the hub' menu link
- add a grid to the home page like other sections


docs/hub/_toctree.yml
```diff
@@ -502,8 +502,6 @@
       title: "Protect AI"
     - local: security-jfrog
       title: "JFrog"
-  - local: agents
-    title: Agents on Hub
```
pcuenca, line 506: Ah, this is out of place in the "Other" section 👍

PR #2228, opened 2026-02-18
Fix link to git-xet in documentation

Erotemic:
Link git git-xet was broken because a "-" should have been an "_"


pcuenca (approved):
Thank you!

PR #2586, opened 2026-06-23
feat: add docs for buckets s3 compatibility api

jgodlew:
Storage Buckets now expose an **S3-compatible API** via a gateway at https://s3.hf.co, so users can reach their buckets with existing S3 tooling (AWS CLI, boto3, s5cmd, etc.). 

This PR documents that feature and its limitations/differences to AWS S3 in a new `Storage Buckets -> S3 Compatibility` page.

---

> [!NOTE]
> **Low Risk**
> Documentation-only changes with no runtime, auth, or data-handling code modifications.
> 
> **Overview**
> Documents the **S3-compatible gateway** (`https://s3.hf.co`) for Storage Buckets so users can use AWS CLI, `boto3`, `s5cmd`, and similar tools without rewriting workflows.
> 
> Adds a new **S3 Compatibility** Hub doc page covering credential generation from access tokens, required client settings (path-style addressing, `us-east-1`, checksum workarounds for newer AWS clients), namespace/bucket addressing patterns, AWS S3 differences/limitations, and an `rclone` import example. The Storage Buckets section in `_toctree.yml` now links this page; **Access Patterns** lists S3 API as an option (replacing the prior “not supported” note), and the main **Storage Buckets** tip points readers to the new guide.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b858ad37f56cfba180dc87b9ff578ac6b16ce775. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/docs/hub/storage-buckets-access.md
+++ b/docs/hub/storage-buckets-access.md
@@ -10,7 +10,7 @@ Beyond the [CLI and Python SDK](./storage-buckets#managing-files), there are sev
 | **Volume mounts** | HF Jobs & Spaces (same idea, managed for you) | [See below](#volume-mounts-in-jobs-and-spaces) |
 | **hf:// paths** (fsspec) | Python data tools (pandas, DuckDB) | [See below](#python-data-tools) |
 | **CLI sync** | Batch transfers, backups | [Sync docs](./storage-buckets#syncing-directories) |
-| **S3 API** | Existing S3 tooling (AWS CLI, boto3, rclone) | [S3-Compatible API](./storage-buckets-s3) |
+| **S3 API** | Existing S3 tooling (AWS CLI, boto3, s5cmd) | [S3-Compatible API](./storage-buckets-s3) |
 
 ## Mount as a Local Filesystem
 
--- a/docs/hub/storage-buckets-access.md
+++ b/docs/hub/storage-buckets-access.md
@@ -10,8 +10,7 @@ Beyond the [CLI and Python SDK](./storage-buckets#managing-files), there are sev
 | **Volume mounts** | HF Jobs & Spaces (same idea, managed for you) | [See below](#volume-mounts-in-jobs-and-spaces) |
 | **hf:// paths** (fsspec) | Python data tools (pandas, DuckDB) | [See below](#python-data-tools) |
 | **CLI sync** | Batch transfers, backups | [Sync docs](./storage-buckets#syncing-directories) |
-
-Access through the S3 API is not currently supported, but is on the roadmap.
+| **S3 API** | Existing S3 tooling (AWS CLI, boto3, rclone) | [S3-Compatible API](./storage-buckets-s3) |
 
 ## Mount as a Local Filesystem
 
--- a/docs/hub/storage-buckets-s3.md
+++ b/docs/hub/storage-buckets-s3.md
@@ -1,6 +1,7 @@
 # S3 Compatibility
 
-Storage Buckets can be accessed through an **S3-compatible API**, letting you use existing S3 tooling — the AWS CLI, `boto3`, `rclone`, `s5cmd`, and most other S3 SDKs — against your buckets without changing your code. Requests go through a gateway at `https://s3.hf.co`.
+Storage Buckets can be accessed through an **S3-compatible API**, letting you use existing S3 tooling — the AWS CLI, `boto3`, `s5cmd`, and most other S3 SDKs — against your buckets without changing your code. 
+Requests go through a gateway service at `https://s3.hf.co`.
 
 > [!NOTE]
 > The S3 API works only with [Storage Buckets](./storage-buckets). It does not expose other Hugging Face repository types (models, datasets, Spaces).
@@ -11,6 +12,12 @@ The gateway authenticates with AWS-style access keys derived from a Hugging Face
 
 1. Go to your [Access Tokens settings](https://huggingface.co/settings/tokens). Create a token with the **Create new token** button if you don't already have one.
 2. Find the token in the list, open its dropdown menu, and choose **Generate S3 credentials**.
+
+   <div class="flex justify-center">
+   <img class="block dark:hidden" src="https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/hub/buckets/bucket-generate-s3-credentials-light.png" alt="Generate S3 credentials option in access token's dropdown menu"/>
+   <img class="hidden dark:block" src="https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/hub/buckets/bucket-generate-s3-credentials-dark.png" alt="Generate S3 credentials option in access token's dropdown menu"/>
+   </div>
+
 3. Copy the generated **access key ID** (prefixed `HFAK…`) and **secret access key** somewhere safe — the secret is shown only once.
 
 The S3 credentials inherit the permissions of the underlying access token, so scope your token to the namespaces and buckets you intend to use.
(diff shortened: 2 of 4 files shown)


docs/hub/storage-buckets-s3.md
```diff
@@ -0,0 +1,130 @@
...
+
+## Limitations and Differences from AWS S3
+
+Because Storage Buckets don't model every S3 concept, some behaviors differ or aren't supported.
+
+### Object downloads
+
+The gateway is currently single-region. To improve download performance, `GetObject` typically responds with an HTTP 302 redirect to the nearest Hugging Face CDN edge rather than serving bytes directly.
+
+Some SDKs don't follow redirects from an S3 endpoint, so the gateway detects clients identifying as `aws-cli`, `botocore` (which covers `boto3`), or `aws-sdk-rust` and proxies the data through itself for them. All other clients (`rclone`, `s5cmd`, `curl`, the AWS Go SDK, etc.) receive the 302 and follow it natively, keeping the gateway out of the data path for faster downloads.
```
julien-c, line 106: > proxies the data through itself 

ouch, this sounds costly 😓

julien-c (approved):
really cool!

docs/hub/storage-buckets-s3.md
jgodlew: Yeah, it is unfortunate that the aws-cli/boto libraries don't follow redirects. We might want to think about integrating this API with the Hugging Face CDN if we find that a significant amount of downloads are being proxied.

docs/hub/storage-buckets-s3.md
```diff
@@ -0,0 +1,130 @@
...
+- `ListObjectsV1` is not supported — use `ListObjectsV2`.
+- Only `/` is allowed as the delimiter.
+
+### Other API differences
+
+- **Object metadata**: arbitrary user metadata (`x-amz-meta-*`) is not stored or returned. `Content-Type` is supported.
+- **Unsupported features**: ACLs, bucket policies, object tagging, object versioning, lifecycle rules, server-side encryption (SSE), and bucket notifications are not supported. Objects are always reported with the `STANDARD` storage class. Related request headers and parameters are accepted but ignored.
+- **CopyObject**: server-side copy works only within a single namespace. Cross-namespace copy and `UploadPartCopy` (copying a part from an existing object into a multipart upload) are not supported.
+- **Conditional requests**: `If-Match` / `If-None-Match` preconditions are honored on `PutObject` and on the copy-source of `CopyObject`, but not on `GetObject`.
+- **Multipart upload expiry**: in-flight multipart uploads that are never completed or aborted are automatically expired and cleaned up after 7 days.
```
Pierrci, line 131: i'm curious - how is that one different? is it in the number of days or something else?

Pierrci (commented):
🔥

XciD (approved):

XciD:
We need to link to this doc in moon-landing when we create the S3 secrets I guess.

pcuenca (commented):
🔥

Wauplin (approved):
Very nice!

Wondering if we should add a real use case example already or if that should rather live in a separate doc. Here is a CC suggestion on how to import data using `rclone`:

````
## Example: import data using `rclone`

[`rclone`](https://rclone.org/) is a convenient way to copy data between two S3-compatible stores, so it's a good fit for moving an existing AWS S3 bucket (or any S3-compatible source) into a Storage Bucket.

The idea is to declare two remotes — your source bucket and the Hugging Face gateway — and let `rclone` stream the objects between them. Add both remotes to `~/.config/rclone/rclone.conf`. The first points at your existing S3 bucket; adjust it to match your source (here, plain AWS S3):

```ini
[aws]
type = s3
provider = AWS
access_key_id = AKIA...
secret_access_key = ...
region = us-east-1
```

The second points at the Hugging Face gateway. As with any other client, scope the endpoint to your [namespace](#addressing-buckets), use `path` addressing, and set large multipart sizes so uploads use as few parts as possible:

```ini
[hf]
type = s3
provider = Other
endpoint = https://s3.hf.co/<namespace>
access_key_id = HFAK...
secret_access_key = ...
region = us-east-1
force_path_style = true
upload_cutoff = 2G
chunk_size = 2G
```

Note: replace the `<namespace>` above with the username or organization your buckets are stored in, and use the [S3 credentials](#generating-s3-credentials) generated from your access token. The destination bucket must already exists under that namespace.

Now, let's copy a source bucket into your Storage Bucket using `rclone`:

```bash
rclone copy aws:my-source-bucket hf:my-bucket --progress
```

`rclone copy` only transfers objects that are missing or changed at the destination, so it's safe to re-run to resume an interrupted import or pick up new objects. To make the destination an exact mirror of the source — deleting objects at the destination that no longer exist at the source — use `rclone sync` instead:

```bash
rclone sync aws:my-source-bucket hf:my-bucket --progress
```

> [!TIP]
> For large imports, add `--transfers` and `--checkers` to raise the concurrency (e.g. `--transfers 16 --checkers 16`), and run `rclone check aws:my-source-bucket hf:my-bucket` afterwards to confirm every object made it across.
````

(can be done in a future PR)

**Comment by julien-c** [comment]
@Wauplin Yes, I was wondering about that as well.

File: docs/hub/storage-buckets-s3.md (hunk 2 above)
Comment (jgodlew) [reply docs/hub/storage-buckets-s3.md:131]: AWS S3 doesn't have an expiration on multipart uploads. It is up to the caller to [remove abandoned multipart uploads](https://docs.aws.amazon.com/AmazonS3/latest/userguide/mpuoverview.html#mpu-process). You can, however, [set a lifecycle policy](https://docs.aws.amazon.com/AmazonS3/latest/userguide/mpu-abort-incomplete-mpu-lifecycle-config.html) on the bucket to give an expiry.

So the main difference is that we enforce a 7 day policy instead of the default AWS S3 behavior of no expiration and exposing an optional lifecycle mechanism.

**Comment by davanstrien** [comment]
> Wondering if we should add a real use case example already or if that should rather live in a separate doc. Here is a CC suggestion on how to import data using `rclone`:


Had Claude Code run a bunch of S3 tooling and it found that the `[hf]` remote needs `list_version = 2`. With `provider = Other`, rclone defaults to ListObjects V1, which the gateway returns a 501 for, so listing inside a bucket fails. Since `rclone copy`/`sync` list the destination to compute the diff, the import would hit that too.

So the `[hf]` remote wants:

```ini
[hf]
type = s3
provider = Other
endpoint = https://s3.hf.co/<namespace>
access_key_id = HFAK...
secret_access_key = ...
region = us-east-1
force_path_style = true
upload_cutoff = 2G
chunk_size = 2G
list_version = 2
```

docs/hub/storage-buckets-access.md
```diff
@@ -10,8 +10,7 @@ Beyond the [CLI and Python SDK](./storage-buckets#managing-files), there are sev
 | **Volume mounts** | HF Jobs & Spaces (same idea, managed for you) | [See below](#volume-mounts-in-jobs-and-spaces) |
 | **hf:// paths** (fsspec) | Python data tools (pandas, DuckDB) | [See below](#python-data-tools) |
 | **CLI sync** | Batch transfers, backups | [Sync docs](./storage-buckets#syncing-directories) |
-
-Access through the S3 API is not currently supported, but is on the roadmap.
+| **S3 API** | Existing S3 tooling (AWS CLI, boto3, s5cmd) | [S3-Compatible API](./storage-buckets-s3) |
```
davanstrien, line 13: 🔥

docs/hub/storage-buckets-s3.md
```diff
@@ -0,0 +1,130 @@
...
+Requests go through a gateway service at `https://s3.hf.co`.
+
+> [!NOTE]
+> The S3 API works only with [Storage Buckets](./storage-buckets). It does not expose other Hugging Face repository types (models, datasets, Spaces).
+
+## Generating S3 Credentials
+
+The gateway authenticates with AWS-style access keys derived from a Hugging Face [User Access Token](./security-tokens).
+
+1. Go to your [Access Tokens settings](https://huggingface.co/settings/tokens). Create a token with the **Create new token** button if you don't already have one.
```
davanstrien: Maybe good to clarify required permissions, i.e. read/write. I assume read is possible if you just want to read via S3? It's discussed a bit below but maybe nice to mention as part of the steps so user can follow along

davanstrien (commented):
very nice! It could be worth adding a longer intergrations supported via S3 section at some point but can iterate on that later and focus on most used to start

jgodlew:
> Wondering if we should add a real use case example already

Yeah, I agree that is a good idea. I've added an `Examples` section to the doc with the suggested `Import data using rclone` as the first example (incorporating the list_version change @davanstrien indicated). I think we can probably add more examples as we find more use-cases.

docs/hub/storage-buckets-s3.md
jgodlew: Sounds good! I've mentioned the permissions relationship between the HF Access Token and the S3 credentials during this step.

Wauplin (approved):
Thank you!

docs/hub/storage-buckets-s3.md
Pierrci: ok thanks!

PR #2692, opened 2026-08-07
agents-cli: reflect the hf-cli skill auto-install

davanstrien:
`huggingface_hub` v1.27.0 changed how the `hf-cli` skill gets installed, and `agents-cli.md` still describes the old flow. The release notes link to this page, so it's the first thing a reader lands on.

https://github.com/huggingface/huggingface_hub/pull/4608
https://github.com/huggingface/huggingface_hub/releases/tag/v1.27.0

What changed here:

- The standalone installer now installs the Skill globally (including the Claude Code location), so that's stated up front, with `--exclude-skill` / `-ExcludeSkill` to opt out.
- "Add the CLI Skill" is reframed as the path for people the installer doesn't cover — pip, Homebrew, opting out, or wanting it in one project.
- The tip said the Skill is "always up to date". That is now the opposite of true: it's generated at install time, and v1.27.0 ships a once-a-day reminder precisely because it drifts. Replaced with how to refresh it (`hf update`, `hf skills update -g`) and how to silence the reminder.
- `agent-traces.md` told readers to run `hf skills add` right after the installer. That step is now redundant, so it links here instead.

The equivalent guidance landed in the `huggingface_hub` guides with #4608; this mirrors it in hub-docs.

Commands verified against v1.27.0, and the `--exclude-skill` / `-ExcludeSkill` flags against the deployed scripts at hf.co/cli/.

Unrelated but adjacent: #2539 is a one-line fix to this same file, approved since June 8 and still unmerged. Different line, no conflict.

### DIFF
--- a/docs/hub/agent-traces.md
+++ b/docs/hub/agent-traces.md
@@ -39,12 +39,11 @@ Buckets are especially useful if you want to keep syncing traces as new sessions
 
 ## Upload your traces
 
-After creating a dataset or bucket, install the `hf` CLI with the [recommended standalone installer](/docs/huggingface_hub/guides/cli#getting-started) and log in. If you want your coding agent to run `hf` commands for you, install the Hugging Face CLI skill with `hf skills add`.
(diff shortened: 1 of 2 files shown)


Pierrci (approved):
thanks, lgtm!

julien-c (approved):

PR #1247, opened 2024-03-14
Add GGUF doc page

mishig25:
See the page: https://moon-ci-docs.huggingface.co/docs/hub/pr_1247/en/gguf

Question: where should I add it in the side bar? 

Screenshot 2024-03-14 at 10 22 16 PM

Issue #1176, opened 2024-01-04
Improve search on papers page


Vipitis:
was directed to open the issue here.

The [papers page](https://huggingface.co/papers) has this search for "arxiv id or title". What really bothers me, is that pasting the full arxiv link (or even pdf link) does not find the paper.

**Current workaround**:
manually delete all characters before and after the arxiv id.

Should be a small change

**Example:**
![not working](https://github.com/huggingface/hub-docs/assets/23384483/a86a1a0e-0e09-4acc-b568-e82442f2ae51)
![working](https://github.com/huggingface/hub-docs/assets/23384483/91e9a4b6-c71d-43ec-a324-ac9c3a86d981)

julien-c:
maybe cc @mishig25!

julien-c:
thanks for reporting @Vipitis can you confirm it works fine now?

Vipitis:
I tested this earlier today. It works with arxiv links, arxiv pdf links and even hf papers links.  Thanks for getting this added this quickly 🤗

PR #2742, opened 2026-08-28
docs(xet): correct shard/xorb format specs and stale client defaults

rajatarya:
Aligns the public Xet specs with **xet-core `v1.6.0`**, which is the ground truth for every change here. Five findings from the xet doc-drift audit ([open drift issues](https://github.com/huggingface-internal/xetcas/issues?q=label%3Axet-doc-drift)) — tracked as **#1045, #1152, #1176, #1247, #1264**.

Each change is described below, so the PR is self-contained for review.

## `docs/xet/shard.md` (#1045)

The involved one — two layout-level drifts, both field-labeling rather than size changes, so **no byte offsets shift**.

- **`CASChunkSequenceEntry`**: the trailing 8-byte `_unused` is actually `flags: u32` + `_unused: u32`. `flags` is load-bearing — it carries `MDB_CHUNK_WITH_GLOBAL_DEDUP_FLAG` (`1 << 31`), which marks a chunk eligible for global dedup. Adds a bitfield table and a rule that readers mask undefined bits rather than compare for equality. Entry stays 48 bytes.
- **`MDBShardFileFooter`**: documented 9 fields; source has 17. The two `_buffer` spans masked six lookup offset/count fields (bytes 24–72) and three accounting totals (bytes 168–192). Footer stays 200 bytes.
- **`cas_info_offset` → `xorb_info_offset`** to match source, with a note recording the old name. The section it points at is still called the CAS Info Section here.
- **New: the three lookup tables.** Keys are *truncated* hashes (first 8 bytes as `u64`) and values are *indices*, not byte offsets — so a match is inconclusive and must be confirmed against the full hash. Also documents per-entry sizes and that a `0` count means the table is absent.
- Added the tables to both layout diagrams, and corrected two consequences: the CAS info section ends at its bookend (not `footer_offset`, now that tables may sit between), and footer deserialization can't read every field as `u64` (the HMAC key is 32 B, `_buffer` is 48 B).

## `docs/xet/xorb.md` (#1176, #1264)

- **#1176** — replaces *"no explicit limit on the number of chunks"* with the **8192-chunk cap** (`MAX_XORB_CHUNKS`), scoped precisely: it's producer-side cut logic in the reference client, **not** in the wire format and **not** server-enforced. This is the reconciliation the issue asks for — internal docs presented 8192 as a hard maximum while this doc said there was no limit; neither was quite right.
- **#1264** — documents the `XorbObjectInfo` footer and the trailing `info_length: u32`, previously **absent entirely**, which left readers no documented way to locate the footer (the old diagram's trailing `...` read as "more chunks"). Adds the reader entry point (`end - 4`, length excludes itself), the fixed-size tail layout, and the **uniqueness nonce** — leading 4 bytes of the 16-byte buffer, excluded from the xorb hash, so identical content need not serialize to identical bytes. Pre-nonce xorbs carry an all-zero buffer and stay valid.

## `docs/xet/api.md` (#1247)

One-line fix: `/v2/shards` falls back to `/v1/shards` on a **`404` or `501`**, not `404` alone. A client built strictly from the old text would treat `501` as fatal, and `501` is classed as permanent so no retry papers over it. The sibling reconstruction note already said "404 or 501", so the two notes in this file disagreed. Also notes that the reference client caches the detected version per session.

## `docs/hub/xet/using-xet-storage.md` (#1152)

Three stale defaults:

| Variable | Was | Now |
|---|---|---|
| `HF_XET_CLIENT_AC_INITIAL_UPLOAD_CONCURRENCY` | `1` | `2` |
| `HF_XET_CLIENT_AC_INITIAL_DOWNLOAD_CONCURRENCY` | `1` | `4` |
| `HF_XET_CLIENT_READ_TIMEOUT` | `120s` | `300s` |

## Verification

- Every claim re-checked against `xet-core v1.6.0` source rather than taken from the issue text. Two issues were stale and corrected in the process: #1264 calls the footer `CasObjectInfo`, a name no longer in source (`XorbObjectInfoV1`), and #1176's original "debug-assert only" framing understates the cap, which is real release-path cut logic.
- Byte-layout diagrams were **generated programmatically**, and offset rulers verified to align with column borders. Footer arithmetic checked to sum to exactly 200 bytes and the chunk entry to 48.
- Diagram alignment before/after: `shard.md` had **9** diagrams whose offset rulers didn't line up with their borders on `main`; this branch has **6**. The 3 fixed are the ones regenerated here; none were introduced.

## Deliberately out of scope

Broader `CAS*` → `Xorb*` terminology alignment across `shard.md` (section headings, `CASChunkSequenceHeader`, bookend prose). Only `cas_info_offset` is renamed, because it's a field name clients bind to; the rest is prose and a larger public-terminology decision.

---

> [!NOTE]
> **Low Risk**
> Changes are documentation only; they correct spec drift and env-var defaults without altering runtime code or deployed behavior.
> 
> **Overview**
> Documentation-only update to match **xet-core v1.6.0** and fix doc drift across Hub env vars and Xet wire-format specs.
> 
> **`shard.md`** expands the shard layout: optional **lookup tables** after the CAS info section; footer documents **`xorb_info_offset`** (renamed from `cas_info_offset`), six lookup offset/count fields, and accounting totals; **`CASChunkSequenceEntry`** now exposes **`flags`** (including global-dedup) instead of an 8-byte `_unused` field; parsing notes clarify CAS info ends at the bookend and footer field types.
> 
> **`xorb.md`** adds the **8192-chunk producer cap** (not wire/server enforced), the **`XorbObjectInfo`** footer with trailing **`info_length`**, and the **uniqueness nonce** behavior relative to the xorb hash.
> 
> **`api.md`** states clients should fall back from `/v2/shards` to `/v1/shards` on **`404` or `501`**, with per-session version caching.
> 
> **`using-xet-storage.md`** updates documented defaults: initial upload concurrency **2**, download **4**, read timeout **300s**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit bc810d64bd554a332fb33743e6c9afbf0cfdfcf2. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/docs/hub/xet/using-xet-storage.md
+++ b/docs/hub/xet/using-xet-storage.md
@@ -155,8 +155,8 @@ By default, `xet-core` uses adaptive concurrency — dynamically adjusting paral
 | Environment Variable | Default | Description |
 |---|---|---|
 | `HF_XET_CLIENT_ENABLE_ADAPTIVE_CONCURRENCY` | `true` | Enable or disable adaptive concurrency control. When disabled, concurrency stays at the initial value. |
-| `HF_XET_CLIENT_AC_INITIAL_UPLOAD_CONCURRENCY` | `1` | Starting number of concurrent upload streams. HP mode: `16`. |
-| `HF_XET_CLIENT_AC_INITIAL_DOWNLOAD_CONCURRENCY` | `1` | Starting number of concurrent download streams. HP mode: `16`. |
+| `HF_XET_CLIENT_AC_INITIAL_UPLOAD_CONCURRENCY` | `2` | Starting number of concurrent upload streams. HP mode: `16`. |
+| `HF_XET_CLIENT_AC_INITIAL_DOWNLOAD_CONCURRENCY` | `4` | Starting number of concurrent download streams. HP mode: `16`. |
 | `HF_XET_CLIENT_AC_MIN_UPLOAD_CONCURRENCY` | `1` | Lower bound for upload concurrency. HP mode: `4`. |
 | `HF_XET_CLIENT_AC_MIN_DOWNLOAD_CONCURRENCY` | `1` | Lower bound for download concurrency. HP mode: `4`. |
 | `HF_XET_CLIENT_AC_MAX_UPLOAD_CONCURRENCY` | `64` | Upper bound for upload concurrency. HP mode: `124`. |
@@ -178,7 +178,7 @@ By default, `xet-core` uses adaptive concurrency — dynamically adjusting paral
 | `HF_XET_CLIENT_RETRY_BASE_DELAY` | `3000ms` | Base delay between retries (with exponential backoff). |
 | `HF_XET_CLIENT_RETRY_MAX_DURATION` | `360s` | Maximum total time to spend retrying a request. |
 | `HF_XET_CLIENT_CONNECT_TIMEOUT` | `60s` | TCP connection timeout. |
-| `HF_XET_CLIENT_READ_TIMEOUT` | `120s` | Read timeout for HTTP responses. |
+| `HF_XET_CLIENT_READ_TIMEOUT` | `300s` | Read timeout for HTTP responses. |
 | `HF_XET_CLIENT_IDLE_CONNECTION_TIMEOUT` | `60s` | Timeout before idle connections are closed. |
 | `HF_XET_CLIENT_MAX_IDLE_CONNECTIONS` | `16` | Maximum number of idle connections in the pool. |
 
--- a/docs/xet/api.md
+++ b/docs/xet/api.md
@@ -229,7 +229,7 @@ A typical successful stream:
 If validation stalls, the server periodically re-emits the last progress event as a heartbeat so the stream never looks dead, which lets clients keep a short read timeout.
 
 > [!NOTE]
-> Additive change: `/v1/shards` is unchanged. Clients SHOULD try `/v2/shards` and fall back to `/v1/shards` on `404 Not Found`.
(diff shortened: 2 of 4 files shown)


rajatarya:
cc @sirahd for review alongside @seanses — GitHub won't accept a formal review request here since you're not a collaborator on this repo, so flagging it this way instead.

assafvayner:
The lookup tables were intentionally omitted from the spec, and I still think they should be.

The lookup tables are never needed in the shard when using the shard format as a data serialization format for upload or the global dedupe API. They are only used by xet-core as an optimized on-disk data structure for fast dedupe lookup during the upload procedure. Is there a reason we want to add them in now?

I'm also not very well versed in the usage and format of the lookup, will need to review, @seanses may know better.

assafvayner:
xorb format footer nonce is a server-side component, it is not trusted from the client and doesn't need to be mentioned.

assafvayner (changes requested):
see other comments. the docs changes and shard api notes are fine.

seanses:
> The lookup tables were intentionally omitted from the spec, and I still think they should be.

Agree with Assaf. The lookup tables are deliberately removed from shards on upload, the server never needs them.

seanses:
> xorb format footer nonce is a server-side component, it is not trusted from the client and doesn't need to be mentioned.

Actually, the entire Xorb footer is also deliberately removed from Xorb on upload, the server reconstructs it from the verified chunk data. So we shouldn't mention it in the wire protocol.

PR #2750, opened 2026-08-31
[Inference Providers] drop the conversational tag requirement

hanouticelina:
The Hub no longer requires the `conversational` tag to register a model as a chat-completion model for inference providers (internal PR: huggingface-internal/moon-landing#19696). New rule: any model with pipeline tag `text-generation` or `image-text-to-text` can be registered as `conversational` — the registered mapping decides, not the model's tags — and the legacy `text-generation` inference task is no longer accepted at registration.

`register-as-a-provider`:
- the `chatCompletion` note and the registration validation rules,
- the `task` field examples for both mapping types,
- the provider helper sections (JS + Python): only `BaseConversationalTask` is relevant for new providers now,
- the mapping listing example and the latency requirements, which still mentioned the `text-generation` task.

`models-widgets`: the conversational widget is now shown on all `text-generation` / `image-text-to-text` models, without requiring the tag (the text-generation widget is removed).

The `conversational` tag itself is unchanged (still derived from the chat template), so model-search links keep using `other=conversational`.

---

> [!NOTE]
> **Low Risk**
> Docs-only change describing registration and widget rules; no runtime or API code in this diff.
> 
> **Overview**
> **Documentation now matches Hub inference-provider behavior:** chat/completion mappings no longer depend on the model’s `conversational` tag, and the legacy `text-generation` registration task is dropped in favor of `conversational` for `text-generation` and `image-text-to-text` models.
> 
> In **`models-widgets.md`**, the conversational widget is described as shown on **all** models with those pipeline tags (not only when tagged `conversational`). **`register-as-a-provider.md`** updates the `chatCompletion` tip, mapping `task` examples, and validation rules; provider helper guidance now points only to **`BaseConversationalTask`** (JS/Python), not `BaseTextGenerationTask`. The sample mappings JSON and latency wording drop the separate `text-generation` task bucket.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b79f0dbc618cc84d4dbf010c010c2083e401082d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/docs/inference-providers/integrations/codex.md
+++ b/docs/inference-providers/integrations/codex.md
@@ -55,7 +55,7 @@ You can also run a one-off command non-interactively:
 codex exec --profile huggingface "Explain what this repository does."
 ```
 
-Replace `openai/gpt-oss-120b` with any model available on [Inference Providers](https://huggingface.co/models?pipeline_tag=text-generation&inference_provider=all&sort=trending). You can append a provider suffix to pin a specific provider (e.g. `openai/gpt-oss-120b:groq`). Setting an explicit provider gives you deterministic routing, while omitting it lets the router fall back to alternative providers for better resilience. You can also append a `:cheapest` or `:fastest` suffix to prefer cheaper or faster providers.
+Replace `openai/gpt-oss-120b` with any model available on [Inference Providers](https://huggingface.co/models?other=conversational&inference_provider=all&sort=trending). You can append a provider suffix to pin a specific provider (e.g. `openai/gpt-oss-120b:groq`). Setting an explicit provider gives you deterministic routing, while omitting it lets the router fall back to alternative providers for better resilience. You can also append a `:cheapest` or `:fastest` suffix to prefer cheaper or faster providers.
 
 To override the model for a single run without editing your config, pass `-m`:
 
@@ -85,4 +85,4 @@ Replace `"your-org-name"` with the name of the organization you want to bill to.
 
 - [Codex Documentation](https://developers.openai.com/codex)
 - [Codex Advanced Configuration](https://developers.openai.com/codex/config-advanced)
-- [Available models on Inference Providers](https://huggingface.co/models?pipeline_tag=text-generation&inference_provider=all&sort=trending)
+- [Available models on Inference Providers](https://huggingface.co/models?other=conversational&inference_provider=all&sort=trending)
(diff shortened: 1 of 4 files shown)


docs/inference-providers/integrations/codex.md
```diff
@@ -55,7 +55,7 @@ You can also run a one-off command non-interactively:
 codex exec --profile huggingface "Explain what this repository does."
 ```
 
-Replace `openai/gpt-oss-120b` with any model available on [Inference Providers](https://huggingface.co/models?other=conversational&inference_provider=all&sort=trending). You can append a provider suffix to pin a specific provider (e.g. `openai/gpt-oss-120b:groq`). Setting an explicit provider gives you deterministic routing, while omitting it lets the router fall back to alternative providers for better resilience. You can also append a `:cheapest` or `:fastest` suffix to prefer cheaper or faster providers.
+Replace `openai/gpt-oss-120b` with any model available on [Inference Providers](https://huggingface.co/models?pipeline_tag=text-generation&inference_provider=all&sort=trending). You can append a provider suffix to pin a specific provider (e.g. `openai/gpt-oss-120b:groq`). Setting an explicit provider gives you deterministic routing, while omitting it lets the router fall back to alternative providers for better resilience. You can also append a `:cheapest` or `:fastest` suffix to prefer cheaper or faster providers.
```
Wauplin: Arf, too bad. The `conversational` tag was very useful to select both `text-generation` and `image-text-to-text` models (and soon `any-to-text`?). I think I prefer to keep `other=conversational` here

Wauplin: We can even auto-tag all models registered as chat-completion by at least 1 Inference Provider as conversational. What do you think?

Wauplin (commented):
A bit hesitant on my proposition but I think it's still better than selecting only text-generation models (we loose models like Kimi K3, GLM 5.3 Flash, Muse, etc.)

docs/inference-providers/integrations/codex.md
hanouticelina: ^ ah yes, good idea

hanouticelina: (Comment by Opus) Good point, reverted — kept `other=conversational` in all the model-search links (here, `index.md` and `models-widgets.md`). The tag is still derived from the chat template and still spans both pipeline tags, it just is not a registration requirement anymore.

The PR now only touches `register-as-a-provider.md`.

Wauplin (approved):
Approved for when https://github.com/huggingface-internal/moon-landing/pull/19696 gets merged!

And let's do the conversational auto-tag if everyone agrees on it

PR #2758, opened 2026-09-03
Clarify Python package reuse with Jobs framework images

davanstrien:
Clarifies which Python packages Jobs uses when running a framework image directly versus running a UV script inside it.

- Explain `hf jobs run` versus `hf jobs uv run` near the top.
- Correct the TRL/PyTorch and vLLM claims: selecting an image does not automatically expose its Python packages to UV scripts with inline dependencies.
- Explain how to reuse the image's prebuilt stack while adding script-specific dependencies, including import precedence and dependency-resolution limitations.
- Document `--` for forwarding conflicting command options, with a tested TRL help example.

**Validation:** CPU Jobs on `huggingface/trl:1.12.0` confirmed direct imports, UV isolation, and reuse through the image interpreter plus `PYTHONPATH`. A separate Job confirmed `-- trl --help` reaches TRL. These checks cover imports and argument forwarding, not GPU training or inference.

Related: https://github.com/huggingface/trl/issues/7028 and https://github.com/huggingface/trl/pull/7054.

### DIFF
--- a/docs/hub/jobs-configuration.md
+++ b/docs/hub/jobs-configuration.md
@@ -44,10 +44,12 @@ Specify the Docker image and the command to run as you would with docker:
 >>> hf jobs run ubuntu echo "Hello from the cloud!"
 ```
 
-All options to Jobs must be provided before the command. A `--` can be used to separate the command from jobs/uv options for clarity, e.g.
+Place Jobs options before the image. If your command uses options that Jobs also recognises,
+such as `--help` or `--timeout`, add `--` between the image and command. Otherwise, Jobs can
+consume those options instead of forwarding them to your command:
 
 ```bash
->>> hf jobs run --token hf_xxx ubuntu -- echo "Hello from the cloud!"
+>>> hf jobs run --flavor cpu-basic huggingface/trl -- trl --help
 ```
 
 Find the list of all arguments in the [CLI documentation](https://huggingface.co/docs/huggingface_hub/package_reference/cli#hf-jobs-run).
--- a/docs/hub/jobs-popular-images.md
+++ b/docs/hub/jobs-popular-images.md
@@ -1,8 +1,17 @@
 # Popular Images
 
-Here is the list of ready-to-use Docker images from popular frameworks that you can use in Jobs with uv.
+Here is the list of ready-to-use Docker images from popular frameworks that you can use in Jobs.
 
-These Docker images already have uv installed but if you want to use an image + uv for an image without uv installed you’ll need to make sure uv is installed first. This will work well in many cases but for LLM inference libraries which can have quite specific requirements, it can be useful to use a specific image that has the library installed.
+Choose how to run your code:
+
+- [`hf jobs run`](./jobs-configuration#docker-jobs) executes a command directly in the image,
+  so you can use its preinstalled Python packages.
+- [`hf jobs uv run`](./jobs-configuration#uv-jobs) runs your script with UV. For a script with
+  a `# /// script` dependency header, UV creates an isolated environment. `--image` supplies
+  the system environment, but does not automatically expose the image's Python packages.
+
+The UV examples below require an image with `uv` installed. To combine UV-managed dependencies
+with the image's preinstalled packages, see [Reuse the image's packages and add dependencies with UV](#reuse-the-images-packages-and-add-dependencies-with-uv).
 
 > [!TIP]
 > For GPU inference libraries, pass `--image` so the run gets a matching CUDA system stack
@@ -21,9 +30,10 @@ Use the `--image` argument to use this Docker image:
 ```
 
 > [!TIP]
-> The `vllm/vllm-openai` image bundles the CUDA toolkit and prebuilt vLLM/FlashInfer kernels.
-> Using it is usually all you need; you can also reuse its prebuilt Python builds — see [Using
-> framework images for GPU libraries](#using-framework-images-for-gpu-libraries) below.
+> With `hf jobs uv run`, this image provides CUDA tooling, but UV resolves vLLM from your
+> script's dependencies rather than using the image's preinstalled vLLM build and its kernels.
+> To use the image's build, run a command with `hf jobs run`, or use the
+> [reuse approach](#reuse-the-images-packages-and-add-dependencies-with-uv) below.
 
 You can find more information on vLLM batch inference on Jobs in [Daniel Van Strien's blog post](https://danielvanstrien.xyz/posts/2025/hf-jobs/vllm-batch-inference.html).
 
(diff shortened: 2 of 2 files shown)


albertvillanova (approved):
Thanks for picking this up so fast. Approving: this fixes the wrong claim, and the trim bullet is the part that turns the mechanism from a shadowed reinstall into an actual reuse. The `PYTHONPATH` shadowing note is the right thing to warn about, since a version pinned on purpose losing to the image is the kind of thing people would otherwise debug for an hour.

One non-blocking correction, which I tested rather than reasoned about. The new TRL paragraph says the bare `--image` form gives "the image's CUDA stack and matched PyTorch build". The second half overstates it.

On an image with `packaging==23.2` installed system wide, running `uv run` on two scripts:

| script header | imported |
| --- | --- |
| `dependencies = ["packaging==24.0"]` | `24.0`, from `/root/.cache/uv/environments-v2/...` |
| `dependencies = []` | `ImportError: No module named 'packaging'` |

The second row is the conclusive one: the environment UV builds does not see the image's site-packages at all, so a package that is present in the image is not importable unless the header asks for it. Without `PYTHONPATH` the script therefore imports the PyTorch UV installed from the header, not the image's build.

The CUDA stack half is right, and it is the real reason to pass the image in that form, since it is what anything compiling against CUDA needs.

Suggested wording: "This gives your script the image's CUDA stack, but not its Python packages, including its TRL and its PyTorch build."

Worth fixing here specifically because it is the same class of overstatement the PR is correcting, one level down.

Separately, and definitely not for this PR: the `3.11` in `/opt/conda/lib/python3.11/site-packages` is decided by the `FROM` line in TRL's `docker/trl/Dockerfile`, and I could not find that version anywhere in the image metadata, so today it is discoverable only by running something. If you want to keep the expert path documented, I can add labels on the TRL side so it can be read straight from the registry without pulling the image:

```dockerfile
LABEL trl.python="/opt/conda/bin/python3"
LABEL trl.site-packages="/opt/conda/lib/python3.11/site-packages"
```

Happy to skip that if you would rather not depend on it.

On the TRL side I have sent https://github.com/huggingface/trl/pull/7054 to fix the equivalent claim in our own Jobs guide. It links here for the image details rather than restating them, so the two should not drift.

albertvillanova:
Following up on my own approval, because I found something bigger than the sentence I flagged, and it affects the vLLM section rather than only TRL. Still not asking to block the merge, this can land as is and be fixed on top.

The distinction that resolves all of it is between the two entry points, which behave in opposite ways:

| entry point | runs the image's Python packages |
| --- | --- |
| `hf jobs run` / `run_job` | yes, the command runs in the image directly |
| `hf jobs uv run` / `run_uv_job` | no, UV resolves the script header into its own environment |

I checked `run_job` in `huggingface_hub`: the command goes straight into the job spec with no UV wrapping. And I tested the `uv run` side rather than reasoning about it, on an image with `packaging==23.2` installed system wide:

| script header | imported |
| --- | --- |
| `dependencies = ["packaging==24.0"]` | `24.0`, from `/root/.cache/uv/environments-v2/...` |
| `dependencies = []` | `ImportError: No module named 'packaging'` |

The second row is conclusive: the environment UV builds does not see the image's site-packages at all.

This page documents `hf jobs uv run` throughout, while describing benefits that only the `hf jobs run` form delivers. So the vLLM tip has the same overstatement as the TRL one, and I think it matters more:

> The `vllm/vllm-openai` image bundles the CUDA toolkit and prebuilt vLLM/FlashInfer kernels. Using it is usually all you need

Under `hf jobs uv run`, vLLM has to be in the script header, and what gets imported is the PyPI build, not the image's prebuilt kernels. vLLM is the case where those kernels matter most, so "usually all you need" points people at the one thing the bare form does not give them.

The page's top tip is already correct, since it scopes the benefit to "a matching CUDA system stack (toolkit, `nvcc`, libraries)". It is the two per-library tips that go further than that.

Suggestions, in whatever form you prefer:

- vLLM tip: keep the CUDA toolkit claim, drop or qualify the prebuilt kernels one, and point at `hf jobs run` for actually running the image's vLLM.
- TRL paragraph: "This gives your script the image's CUDA stack, but not its Python packages, including its TRL and its PyTorch build."
- Possibly a short note near the top naming the two entry points, since every per-library section inherits this.

For reference, `trl-jobs` uses the `run` form for exactly this reason, calling `run_job(image="huggingface/trl", command=["trl", "sft", ...])`.

I have documented both entry points on the TRL side in https://github.com/huggingface/trl/pull/7054, which links here rather than restating the flags and paths.

albertvillanova:
Short addendum: I had the `hf jobs run` row of that table from reading `run_job` in `huggingface_hub`, not from running anything. It is now tested too, so both rows are empirical rather than one of each.

```
$ hf jobs run --flavor cpu-basic huggingface/trl python -c "import sys, trl; print('TRL_VERSION', trl.__version__); print('ARGV', sys.argv)" --model_name_or_path Qwen/Qwen2-0.5B-Instruct
TRL_VERSION 1.12.0
ARGV ['-c', '--model_name_or_path', 'Qwen/Qwen2-0.5B-Instruct']
```

`TRL_VERSION` is the TRL installed in the image, so the `run` form really does execute the image's own build, and it matches the tag the image was published under.

That is the part that matters for the vLLM tip. "Prebuilt vLLM/FlashInfer kernels" is exactly what the `run` form delivers and what the `uv run` form does not, so the claim is not wrong so much as attached to the wrong command.

One sharp edge that turned up on the way, worth a line somewhere in this page if you agree. `hf jobs run` is declared with `ignore_unknown_options` but not `allow_interspersed_args=False`, so options the wrapper recognises are consumed wherever they appear:

```
$ hf jobs run --flavor cpu-basic huggingface/trl trl --help
# prints hf jobs run's own help, not TRL's
```

Unknown flags do pass through untouched, as the `ARGV` line above shows, so this only bites on names that collide with `-e`, `-s`, `-l`, `-v`, `-d`, `--timeout`, `--namespace`, `--token`, `--flavor` or `--help`.

davanstrien:
Thanks @albertvillanova I just confirmed this on `huggingface/trl:1.12.0` with three CPU Jobs:

- [Direct execution](https://huggingface.co/jobs/davanstrien/6a9a94d6259f8e97255ddcef) imports the image's TRL and PyTorch.
- [UV with an empty dependency header](https://huggingface.co/jobs/davanstrien/6a9a94dce686246ca69a0e6b) imports neither.
- [UV with the image interpreter and PYTHONPATH](https://huggingface.co/jobs/davanstrien/6a9a94df259f8e97255ddcf1) restores both imports from the image's site-packages.

Updated the page to distinguish the two entry points, correct the TRL/PyTorch and vLLM wording, and explain the reuse use case: keeping the image's prebuilt stack while adding script-specific dependencies with UV. Kept the import-precedence warning and clarified that UV still resolves dependencies independently.

Also checked the flag collision you flagged: without a separator, `trl --help` shows the Jobs wrapper's help; with `hf jobs run --flavor cpu-basic huggingface/trl:1.12.0 -- trl --help`, it reaches TRL and prints its help ([test job](https://huggingface.co/jobs/davanstrien/6a9a975d259f8e97255ddd8b)). Added an explicit explanation and example to the Docker Jobs configuration section, with a linked tip from Popular Images. This is a bit of annoying gotcha still so maybe at some point it can be smoothed out a bit more. 

I think your TRL PR already covers the corresponding entry-point distinction.

lhoestq (approved):
lgtm !

PR #2212, opened 2026-02-09
[Datasets] Add pyiceberg docs

lhoestq:
New:
* PyIceberg docs (preview: https://moon-ci-docs.huggingface.co/docs/hub/pr_2212/en/datasets-pyiceberg)

Updated:
* DuckDB docs (preview: https://moon-ci-docs.huggingface.co/docs/hub/pr_2212/en/datasets-duckdb)

cc @kszucs

---

> [!NOTE]
> **Low Risk**
> Low risk: documentation-only changes, but adds new guidance and examples that could confuse users if commands/URIs are inaccurate.
> 
> **Overview**
> **Adds new `PyIceberg` documentation** describing how to use `faceberg` to deploy an Iceberg datasets catalog on Spaces, authenticate, add Hub datasets, and query them via `faceberg` or native `pyiceberg` REST catalog clients.
> 
> **Updates `DuckDB` docs** to include a dedicated authentication section, refreshed `hf://` querying examples/output, and a new section showing how to query datasets through an Iceberg catalog (`faceberg.namespace.dataset`) including split-based filtering and catalog deployment steps.
> 
> **Navigation/index updates**: adds `PyIceberg` (and `PyArrow`) entries to the datasets libraries toctree and lists `PyIceberg` in the libraries table.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6337ea48f847db74a0c2561be6d0d74ffe5fb20d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #2768, opened 2026-09-07
Document native hf://buckets/ support in DuckDB (2.0+)

davanstrien:
DuckDB's `httpfs` extension now parses `hf://buckets/` paths directly (duckdb/duckdb-httpfs#293, merged 2026-09-07). It ships with DuckDB 2.0, which is not released yet, so this PR stays a draft until it is.

## Changes

- **storage-buckets-access**: new "Query with DuckDB" section with the native route first, and the DuckDB 1.x fallbacks (`HfFileSystem` from Python, S3 gateway elsewhere) in a note. The overview table gets a DuckDB row, and the DuckDB snippet leaves "Python Data Tools", which is now pandas-only.
- **storage-buckets-integrations**: DuckDB entry next to the other libraries; removed from "Coming soon".
- **datasets-duckdb**: the Storage Buckets tip leads with the native query.
- **storage-buckets-s3**: one-sentence pointer to the native route. The gateway recipe stays, since it is still the CLI route on 1.x and the only way to write.

## Verified

Ran on an HF Job against a CI build of DuckDB `c38d7e0739` (`v2.0.0-dev`) with the matching httpfs binary from the #293 pipeline:

- public bucket read with `**/*.parquet` glob
- private bucket read via `CREATE SECRET (TYPE huggingface, PROVIDER credential_chain)` with `HF_TOKEN`
- private bucket without a secret fails
- `@revision` on a bucket path is rejected ("buckets do not support revisions")
- `COPY ... TO 'hf://buckets/...'` is not supported yet (`DirectoryExists is not implemented`), so the docs say read-only for now

## Before marking ready

- [ ] DuckDB 2.0 released; confirm the version number in the text
- [ ] re-run the checks above on the released build
- [ ] rebase if #2212 (pyiceberg, touches `datasets-duckdb.md`) has merged

cc @cfahlgren1 @lhoestq

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/docs/hub/storage-buckets-s3.md
+++ b/docs/hub/storage-buckets-s3.md
@@ -163,7 +163,9 @@ s3.download_file("my-bucket", "models/model.safetensors", "model.safetensors")
 
 ### Query a bucket with DuckDB
 
-With the `httpfs` extension, [DuckDB](https://duckdb.org/) can read Parquet (and other formats) straight from a bucket over the S3 gateway. If you are on DuckDB 2.0 or later, `httpfs` also reads `hf://buckets/` paths directly with your Hugging Face token and no S3 secret, see [Query with DuckDB](./storage-buckets-access#query-with-duckdb).
+On DuckDB 2.0 or later, the `httpfs` extension reads `hf://buckets/` paths directly. See [Query with DuckDB](./storage-buckets-access#query-with-duckdb).
+
+For earlier versions of DuckDB, or to write query results back to a bucket with `COPY ... TO`, go through the S3 gateway instead:
 
 ```sql
 INSTALL httpfs;
--- a/docs/hub/storage-buckets-s3.md
+++ b/docs/hub/storage-buckets-s3.md
@@ -163,7 +163,7 @@ s3.download_file("my-bucket", "models/model.safetensors", "model.safetensors")
 
 ### Query a bucket with DuckDB
 
-With the `httpfs` extension, [DuckDB](https://duckdb.org/) can read Parquet (and other formats) straight from a bucket:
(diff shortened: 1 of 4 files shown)


docs/hub/storage-buckets-s3.md
```diff
@@ -163,7 +163,7 @@ s3.download_file("my-bucket", "models/model.safetensors", "model.safetensors")
 
 ### Query a bucket with DuckDB
 
-With the `httpfs` extension, [DuckDB](https://duckdb.org/) can read Parquet (and other formats) straight from a bucket:
+With the `httpfs` extension, [DuckDB](https://duckdb.org/) can read Parquet (and other formats) straight from a bucket over the S3 gateway. If you are on DuckDB 2.0 or later, `httpfs` also reads `hf://buckets/` paths directly with your Hugging Face token and no S3 secret, see [Query with DuckDB](./storage-buckets-access#query-with-duckdb).
```
julien-c: maybe reverse the two sentences

davanstrien: updated to put the native path first

docs/hub/storage-buckets-s3.md
```diff
@@ -163,7 +163,9 @@ s3.download_file("my-bucket", "models/model.safetensors", "model.safetensors")
 
 ### Query a bucket with DuckDB
 
-With the `httpfs` extension, [DuckDB](https://duckdb.org/) can read Parquet (and other formats) straight from a bucket:
+On DuckDB 2.0 or later, the `httpfs` extension reads `hf://buckets/` paths directly. See [Query with DuckDB](./storage-buckets-access#query-with-duckdb).
```
lhoestq, line 166: nit: once httpfs is out we can put an example here, otherwise visually the S3 gateway alternative is more prominent and people might just copy paste the code there

edit: ah my bad it's in the s3 docs, it's fine then !

PR #2788, opened 2026-09-17
Add Pixeltable to Datasets Libraries documentation

pierrebrunelle:
### Summary
This PR adds documentation for [Pixeltable](https://github.com/pixeltable/pixeltable) under the Datasets Libraries section.

Pixeltable is an open-source declarative multimodal AI data engine for tables, computed columns, model inference, incremental computation, and vector search. It features built-in integration with Hugging Face Hub datasets (reading Parquet/CSV via `hf://`), native execution of Hugging Face Transformers and Sentence Transformers models as computed columns (`pixeltable.functions.huggingface`), and native vector indexing.

### Changes
- Added `docs/hub/datasets-pixeltable.md` with quickstart, Hub data loading, computed columns with Hugging Face models, and vector search.
- Added Pixeltable to `docs/hub/datasets-libraries.md` table.
- Added `datasets-pixeltable` entry to `docs/hub/_toctree.yml`.


davanstrien (changes requested):
Thanks for the PR! I had an agent try the examples against pixeltable 0.7.8, and it seemed some don't run as written:

- `pip install "pixeltable[transformers]"` — no such extra (`uv pip compile` warns `does not have an extra named transformers`)
- `pxt.create_table("hf_dataset", df)` — second positional arg is `schema`, needs `source=df`
- `from pixeltable.functions.huggingface import pipeline` — no `pipeline` UDF; `text_classification.using(model_id=...)` looks like the equivalent
- `t.order_by(sim, desc=True)` — signature is `asc=True`, so `asc=False`

Could you do a second pass on the examples to make sure they all work okay? 
For "Push to Hub", I couldn't find a Hub-aware export, so that column may need a ❌ unless I/agent missed something?

PR #2801, opened 2026-09-21
Add a webhook guide for processing new bucket files with Jobs

davanstrien:
Adds a webhook guide for a pattern that buckets, webhooks and Jobs now support together: when files land in a bucket, a webhook starts a Job that processes only the changed files, with no server to host.

The guide explains the pattern first (what the bucket payload gives the Job, why the output goes to a second bucket), then sets it up end to end with one example: CSV / JSON uploads converted to optimized Parquet by a ready-made script, [`uv-scripts/data-processing/optimize-parquet.py`](https://huggingface.co/datasets/uv-scripts/data-processing/blob/main/optimize-parquet.py). It ends with other uses of the same setup and when a scheduled Job is the better fit.

Also in this PR:

- `jobs-webhooks.md`: add a bucket example, `bucket` as a repo type, `headSha` not sent for buckets, `updatedFiles`, and a note that webhook runs don't keep the Job's volumes (so UV scripts run from a URL with `hf jobs run ... uv run <url>`).
- Link the guide from `webhooks.md` and the toctree, next to the other webhook guides.

The pipeline was tested end to end on Jobs (`cpu-upgrade`): one upload → one Job (~15 s) → Parquet with content-defined chunking and a page index in the output bucket. I'll re-run the exact commands from the guide once #4958 is released.

**Draft until** huggingface/huggingface_hub#4958 is released: the guide uses `hf webhooks create --watch bucket:...`.


davanstrien:
cc @lhoestq for viz (will wait for next hf release to merge)

